PR #2, opened 2016-07-19
Adding functions to upload files to a dataset

calfonso:


kjaym:
We should remove all of the old search features from this code... They are completely deprecated in the new matrix sites.

PR #23, opened 2016-10-13
added pif and file versioning retrieval methods

mcgalcode:


README.md
```diff
@@ -12,7 +12,31 @@ This package is an implementation of the Citrination API.
...
 In order to interact with older versions of Citrination, please clone
 the legacy branch on this repo.
+
+
+### Troubleshooting
+
+It is possible that you will run into problems if you are using an older version of OpenSSL and running MacOSX. If the following error happens when using the requests library to retrieve information from citrination:
+
+```
+requests.exceptions.SSLError: ("bad handshake: Error([('SSL routines', 'SSL23_GET_SERVER_HELLO', 'sslv3 alert handshake failure')],)",)
```
kjaym: This is great! Thanks for adding it.

citrination_client/client.py
```diff
@@ -67,6 +67,32 @@ def upload_file(self, file_path, data_set_id):
         else:
             return None
 
+    def show_latest_dataset_files(self, dataset_id):
```
kjaym: For all of these, can you add documentation in the format of the the upload_file method. This library is user-facing so it's especially important that it be documented.

citrination_client/client.py
```diff
@@ -67,6 +67,32 @@ def upload_file(self, file_path, data_set_id):
...
+        return self._get_content_from_url(self.api_url + '/data_sets/' + str(dataset_id) + '/latest')
+
+    def show_dataset_files(self, dataset_id):
```
kjaym: What is the difference between this and the previous method, show_latest_dataset_files?

citrination_client/client.py
```diff
@@ -67,6 +67,32 @@ def upload_file(self, file_path, data_set_id):
...
+        return self._get_content_from_url(self.api_url + '/data_sets/' + str(dataset_id) + '/files')
+
+    def show_latest_dataset_file(self, dataset_id, file_path):
```
kjaym: For each of these can you do the following:

- Change "show_..." to "get_..." - I think that's more descriptive of the fact that you are retrieving the record.
- Combine e.g. "show_pif_version" and "show_pif_latest" into a single method "get_pif(self, dataset_id, uid, version=None)". A few important things here... (1) uid rather than pidf_uuid is more consistent with the labels that users use on upload (2) version = None should retrieve the latest version, version != None should get the specified version. Do this same thing for all methods that have a latest and versioned type.

PR #27, opened 2016-10-21
Adding name, description, sharing options to create and update dataset

calfonso:


kjaym:
@calfonso Can you bump the version of this library in setup.py. Then add that version to the cli dependency.

citrination_client/client.py
```diff
@@ -119,7 +119,7 @@ def _get_content_from_url(self, url):
 
         :param url: The URL to make the GET request to
         :return: The response object, or an error message object if the request failed
-        """        
+        """
         print(str(url))
```
kjaym: why are you printing the url here?

citrination_client/client.py
```diff
@@ -173,6 +183,33 @@ def _get_create_data_set_url(self):
         """
         return self.api_url+'/data_sets/create_dataset'
 
+    def update_data_set(self, data_set_id, name=None, description=None, share=None):
+        """
+        Update a data set.
+        :param data_set_id:
+        :param name: name of the dataset
+        :param description: description for the dataset
+        :param share: share the dataset with everyone on the site. Valid values are '1' or '0'.
```
kjaym: Can you be very explicit here that 1 means that it is public and 0 means that it is not.

citrination_client/client.py
calfonso: I didn't add this. but happy to remove it.

PR #35, opened 2016-11-11
Small patch to allow python 3 compatibility

montoyjh:


citrination_client/citrination_client.py
```diff
@@ -20,7 +20,7 @@ def __init__(self, api_key, site='https://citrination.com'):
         the STEEL site on citrination, use 'https://STEEL.citrination.com'.
         :type site: String
         """
-        self.headers = {'X-API-Key': urllib.quote(api_key), 'Content-Type': 'application/json'}
+        self.headers = {'X-API-Key': urllib.parse.quote(api_key), 'Content-Type': 'application/json'}
```
kjaym: Thanks for pointing this out! I believe that this will actually break python 2.x compatibility though. I made an update on the master branch that fixes this https://github.com/kjaym/python-citrination-client/commit/d2855f19f2cdcf5696d65945ef953f0e400b6ee9. See quote_finder.py and how citrination_client.py uses it. Would you mind updating this PR to use that convention?

montoyjh:
Just added those changes.

kjaym:
Thanks very much

PR #43, opened 2016-11-30
check for no hits when iterating through results

kjaym:
@maxhutch This fixes a bug that was happening when no search hits were returned in the iteration. In that case `partial_results.hits` was equal to None and an exception was thrown when calling hits.extend(partial_results.hits). 

I also threw the sleep into its own little if/else statement so that it doesn't sleep when no hits are returned (or on the last iteration loop).


maxhutch:
LGTM.  We should add a test to catch a regression to this :-P

PR #55, opened 2017-01-21
count pifs support in dataset queries

kjaym:


maxhutch:
FYI: you missed a version bump.  I fixed in master.

PR #56, opened 2017-01-24
first pass at simple chemical query

kjaym:
@maxhutch This adds the `simple_chemical_search` method that lets us generate quick, common use case queries without having to build up entire PifQuery objects.

### DIFF
--- a/citrination_client/client.py
+++ b/citrination_client/client.py
@@ -1,11 +1,21 @@
 import json
 import os
+import numbers
 import requests
 from copy import deepcopy
+from six import string_types
 from time import sleep
 from pypif import pif
 from pypif.util.case import keys_to_snake_case
 from citrination_client.util.quote_finder import quote
+from citrination_client.search.pif.query.pif_query import PifQuery
+from citrination_client.search.pif.query.core.field_operation import FieldOperation
+from citrination_client.search.pif.query.core.filter import Filter
+from citrination_client.search.pif.query.core.property_query import PropertyQuery
+from citrination_client.search.pif.query.core.reference_query import ReferenceQuery
+from citrination_client.search.pif.query.core.system_query import SystemQuery
+from citrination_client.search.pif.query.chemical.chemical_field_operation import  ChemicalFieldOperation
+from citrination_client.search.pif.query.chemical.chemical_filter import ChemicalFilter
 from citrination_client.search.pif.result.pif_search_result import PifSearchResult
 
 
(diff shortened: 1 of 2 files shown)


citrination_client/client.py
```diff
@@ -55,6 +65,116 @@ def search(self, pif_query):
...
+            system_query.chemical_formula = \
+                [ChemicalFieldOperation(extract_as='chemical_formula', filter=ChemicalFilter(equal=chemical_formula))]
+        else:
+            system_query.chemical_formula = \
+                [ChemicalFieldOperation(extract_as='chemical_formula')]
+        
+        # Generate the name query for the property
+        if isinstance(property_name, list):
+            property_name_query = \
+                [FieldOperation(extract_as='property_name', filter=Filter(equal=i)) for i in property_name]
```
maxhutch: Shouldn't this be:
```
property_name_query = FieldOperation(extract_as='property_name', filter=[Filter(equal=i) for i in property_name])
```

citrination_client/client.py
```diff
@@ -55,6 +65,116 @@ def search(self, pif_query):
...
+        """
+        system_query = SystemQuery()
+
+        # Add the name operation
+        if isinstance(name, list):
+            system_query.names = [FieldOperation(extract_as='name', filter=Filter(equal=i)) for i in name]
+        elif isinstance(name, string_types):
+            system_query.names = [FieldOperation(extract_as='name', filter=Filter(equal=name))]
+        else:
+            system_query.names = [FieldOperation(extract_as='name')]
```
sinistral: This pattern of 
```
if list, then then return list of reshaped items
else if string then return list of reshaped strings
else return reshaped thing
```
seems to repeat through the rest of the this method.  Extracting it as a function would reduce both the duplication and the obfuscation of the high-level "recipe" that results from the `if/elif/else` blocks.

citrination_client/client.py
kjaym: Both of those work just fine, actually. It looks like that page in the documentation is off. The property name can accept a list of FieldOperations (https://github.com/CitrineInformatics/python-citrination-client/blob/master/citrination_client/search/pif/query/core/property_query.py#L19). 

However, now that you mention it, even though both queries are allowed, the one that you wrote would be more efficient to execute! I am going to rewrite these blocks.

kjaym:
@sinistral This method is really just a builder for the underlying queries that we use. It does not map directly to any parts of the API. Having a high level API though would be a nice idea if this resonates with people! It would be easy to swap out the code in the method with a call to that _future_ API.

citrination_client/client.py
maxhutch: It wasn't working for me.  When I passed in two property names (for the same property), I got no results back.  When I removed either of the names, to get down to a list with a single member, I got results again.

sinistral:
Is there a way to track which of the client methods are actually being used (e.g. a header) so that we can measure which pieces of high-level functionality are seeing use relative to the underlying building blocks?

kjaym:
That is a very good idea, but one we are not doing now!

citrination_client/client.py
kjaym: Okay, I know exactly what you were seeing. We can talk about that later, but it is currently intended behavior. Either way, the changes I am making will negate that issue for this PR.

kjaym:
@maxhutch @sinistral Just pushed code that simplifies the query building logic.

citrination_client/client.py
```diff
@@ -55,6 +63,94 @@ def search(self, pif_query):
...
+        """
+        Helper method that wraps values in a list. If the input is a list then it is returned. If the input is None
+        then an empty list is returned. For anything else, the input value is wrapped as a single-element list.
+
+        :param values: Value to make sure exists in a list.
+        :return: List with the input values.
+        """
+        if values is None:
+            return []
+        elif isinstance(values, list):
```
sinistral: Is a `tuple` also be a legal input sequence, i.e. `simple_chemical_search(name=('A', 'B'))`

sinistral:
Simplified logic is simple.

citrination_client/client.py
kjaym: Not at the moment. That is consistent with the rest of this and our related python tools though.

maxhutch:
LGTM

kjaym:
@sinistral Any other comments?

PR #58, opened 2017-01-24
Adding predict call

maxhutch:
Still debugging.

CC: @sparadiso @jsomara

### DIFF
--- a/citrination_client/client.py
+++ b/citrination_client/client.py
@@ -156,11 +156,11 @@ class CitrinationClient(object):
         Predict endpoint
         :param model_name: The model path
         :param candidates: A list of candidates
-        :return:
+        :return: list of predicted candidates as a map {property: [value, uncertainty]}
         """
 
         # If a single candidate is passed, wrap in a list for the user
-        if type(candidates) is not list:
+        if not isinstance(candidates, list):
             candidates = [candidates]
 
         url = self._get_predict_url(model_name)
--- a/citrination_client/client.py
+++ b/citrination_client/client.py
@@ -55,6 +55,17 @@ class CitrinationClient(object):
             raise RuntimeError('Received ' + str(response.status_code) + ' response: ' + str(response.reason))
         return PifSearchResult(**keys_to_snake_case(response.json()['results']))
 
+    def predict(self, model_name, candidates):
+        url = self._get_predict_url(model_name)
+        body = pif.dumps({"predictionSource": "scalar", "usePrior": True, "candidates": candidates})
+        print("Posting to {} with data={}".format(url, body))
+        response = requests.post(url, data=body, headers=self.headers)
+        return response
+
+    def _get_predict_url(self, model_name):
+        return self.api_url + '/csv_to_models/' + model_name + '/predict'
+
+
     def upload_file(self, file_path, data_set_id, root_path=None):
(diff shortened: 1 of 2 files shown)


maxhutch:
Ready for a look @kjaym 

Huge thanks to @sparadiso and @jsomara

citrination_client/client.py
```diff
@@ -151,6 +151,31 @@ def _get_list(values):
...
+
+        url = self._get_predict_url(model_name)
+        body = pif.dumps(
+            {"predictionRequest": {"predictionSource": "scalar", "usePrior": True, "candidates": candidates}}
+        )
+
+        response = requests.post(url, data=body, headers=self.headers)
+        if response.status_code == requests.codes.ok:
+            return response.json()['candidates']
+        return None
```
kjaym: Raise an exception here?

citrination_client/client.py
```diff
@@ -151,6 +151,31 @@ def _get_list(values):
...
+
+    def _get_predict_url(self, model_name):
+        return self.api_url + '/csv_to_models/' + model_name + '/predict'
```
kjaym: The predict api is only exposed for csv to models?

citrination_client/client.py
```diff
@@ -151,6 +151,31 @@ def _get_list(values):
...
+            candidates = [candidates]
+
+        url = self._get_predict_url(model_name)
+        body = pif.dumps(
+            {"predictionRequest": {"predictionSource": "scalar", "usePrior": True, "candidates": candidates}}
+        )
+
+        response = requests.post(url, data=body, headers=self.headers)
+        if response.status_code == requests.codes.ok:
+            return response.json()['candidates']
```
kjaym: Why not return the entire response?

sparadiso: Do our users care about an HTTP response or the prediction results? I don't really know what your plans were for this, but I would expect the data, not a Response object.

citrination_client/client.py
sparadiso: Yep - it will be exposed for views once those are live.

citrination_client/client.py
sparadiso: Good point. What exception would you raise here? RuntimeError custom?

kjaym: Runtime is fine

citrination_client/client.py
kjaym: Will there be one route? Or is there going to a different route for each method used to generate a model? That is, are we going to have /models/model_name/predict to cover all models, or are we going to have /csv_to_models/model_name/predict and /views/model_name/predict?

citrination_client/client.py
kjaym: I meant `return response.json()`, not `return response`. Sorry, that was not clear.

sparadiso:
Ah, I see. I was just unpacking the only key for them, but if we update
that api to return more information then this will hide it until we update
the client. I see the wisdom in returning the full response in that case.

citrination_client/client.py
maxhutch: Done

citrination_client/client.py
maxhutch: Done

PR #63, opened 2017-01-31
Refactoring test generation to make apidoc easier

maxhutch:
Custom doc files go in src.  When docs are made, they are copied
to source and sphinx-apidoc adds missing docs.


docs/src/conf.py
```diff
@@ -351,3 +351,11 @@
 
 # Example configuration for intersphinx: refer to the Python standard library.
 intersphinx_mapping = {'https://docs.python.org/': None}
+
+def skip(app, what, name, obj, skip, options):
+    if name == "__init__":
+        return False
```
sinistral: Does this mean that initialisers won't be documented?

maxhutch:
The opposite: that function defines which members are "skipped".  Sphinx
skips any "special" functions (starting with "_") by default, so this code
is needed to turn on docs of constructors.

sinistral:
Of course!  I was being dense.  Apologies for the noise.

PR #65, opened 2017-02-05
Deprecate FieldOperation in favor of FieldQuery and version to 1.3.0

kjaym:
@maxhutch This looks like a lot, but it just deprecates FieldOperation and ChemicalFieldOperation in favor of FieldQuery and ChemicalFieldQuery. This is more consistent with the naming throughout the rest of the query language. 

Backwards compatibility is maintained through e.g. https://github.com/CitrineInformatics/python-citrination-client/blob/d98a517a8b5fe588ee2e6f33fc3f895af0d02494/citrination_client/search/pif/query/core/field_operation.py. 

Are we we merging into develop or master? Let me know if it's the former and I will close this and open a new PR.


maxhutch:
Looks good.  We're merging into develop, but I went ahead and fixed that up.  I'll set develop to the default to make it clear.

PR #67, opened 2017-02-09
Feature/multi search

kjaym:
@maxhutch Adds support for multi-search, which is coming soon.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,7 @@
 from setuptools import setup, find_packages
 
 setup(name='citrination-client',
-      version='1.4.0',
+      version='1.2.12',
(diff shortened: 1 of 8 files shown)


setup.py
```diff
@@ -1,7 +1,7 @@
 from setuptools import setup, find_packages
 
 setup(name='citrination-client',
-      version='1.2.12',
+      version='1.4.0',
```
maxhutch: We're bumping versions in release branches now.

maxhutch:
This isn't so much related to this PR in particular, but can we start adding tests with new functionality?

kjaym:
Like we talked about, the functionality is not ready to tests. Adding an issue.

maxhutch:
LGTM

PR #69, opened 2017-02-09
We don't want an old version of pypif

maxhutch:


setup.py
```diff
@@ -7,6 +7,6 @@
       packages=find_packages(exclude=('docs')),
       install_requires=[
           'requests==2.10.0',
-          'pypif==1.0.24',
+          'pypif',
```
kjaym: This is generally not recommended in java/maven world because builds are not necessarily reproducible. Are the conventions different in python?

maxhutch: > It is not considered best practice to use install_requires to pin dependencies to specific versions, or to specify sub-dependencies (i.e. dependencies of your dependencies). This is overly-restrictive, and prevents the user from gaining the benefit of dependency upgrades.

From https://packaging.python.org/requirements/#install-requires

If we stick to semantic versioning, we can specify upper and lower bounds to reflect sensitivity to the API.

maxhutch: > Whereas install_requires requirements are minimal, requirements files often contain an exhaustive listing of pinned versions for the purpose of achieving repeatable installations of a complete environment.

PR #74, opened 2017-03-10
Update file manipulation (DO NOT MERGE UNTIL DECISION IS MADE)

mcgalcode:
Adds upload functionality enhanced with the ability to specify source path (locally) and destination path (in dataset).

Adds ability to retrieve pre-signed download URLs for a file,a batch of files, or "directory" using regex matching.

Adds move and  copy functionality which allows the user to move or copy files or directories within a dataset.


maxhutch:
What does this mean? 
> Adds retrieval (not full download, just presigned URL) functionality for getting a file, batch of files, or "directory" using regex matching.

citrination_client/client.py
```diff
@@ -191,43 +191,41 @@ def predict(self, model_name, candidates):
...
-        :return: Response object or return code if the file was not uploaded.
+        Upload a file, specifying source and dest paths a file (acts as the scp command)
+        :param source_path: The path to the file on the source host
+        :param dest_path: The path to the file where the contents of the transfer will be written (on the dest host)
+        :return: The response object or a boolean False if the file was not uploaded
         """
-        if os.path.isdir(str(file_path)):
-            for path, subdirs, files in os.walk(file_path):
+        source_path = str(source_path)
+        if dest_path == None:
```
maxhutch: Could be `if not dest_path`

citrination_client/client.py
```diff
@@ -191,43 +191,41 @@ def predict(self, model_name, candidates):
...
+
+        if os.path.isdir(source_path):
+            for path, subdirs, files in os.walk(source_path):
                 for name in files:
-                    if root_path:
-                       root = root_path
-                    else:
-                       root = str(file_path)
-                    success = self.upload_file(os.path.join(path, name), data_set_id, root)
+                    success = self.transfer(dataset_id, os.path.join(path, name))
```
maxhutch: Does this discard `dest_path`?

citrination_client/client.py
```diff
@@ -239,6 +237,57 @@ def upload_file(self, file_path, data_set_id, root_path=None):
...
+        :param data_set_id: The dataset id to upload the file to.
+        :return: Response object or return code if the file was not uploaded.
+        """
+        return self.transfer(dataset_id, file_path)
+
+    def move_or_copy(self, dataset_id, move_or_copy, original_path, target_path, is_dir=False):
+        """
+        Move a file within the "filesystem" of a dataset on Citrination.
+        :param dataset_id: The ID of the dataset in which the move will occur
+        :param original_path: The path (to a file or to a directory) to move
```
maxhutch: What are valid values for `move_or_copy`?  What happens if someone specifies a different value?

citrination_client/client.py
```diff
@@ -247,10 +296,8 @@ def get_dataset_files(self, dataset_id, latest = False):
         :param latest: A boolean flag indicating that results should be limited to reporting files from the latest dataset version
         :return: The response object, or an error message object if the request failed
         """
-        if latest == False:
-            return self._get_content_from_url(self.api_url + '/data_sets/' + str(dataset_id) + '/files')
-        elif latest == True:
-            return self._get_content_from_url(self.api_url + '/data_sets/' + str(dataset_id) + '/latest')
+        if latest == False or latest == True:
```
maxhutch: ?!?

mcgalcode: hahaha oh boy

mcgalcode: Oh I see why I did this

citrination_client/client.py
mcgalcode: this was just wrong before. I think I gitt-ed badly.

citrination_client/client.py
```diff
@@ -191,43 +191,42 @@ def predict(self, model_name, candidates):
...
-            url = self._get_upload_url(data_set_id)
-            file_data = {"file_path": str(file_path)}
-            if root_path:
-                file_data["root_path"] = root_path
+            dest_path = str(dest_path)
 
+        if os.path.isdir(source_path):
+            for path, subdirs, files in os.walk(source_path):
+                for name in files:
+                    current_dest_path = os.path.join(dest_path, "/".join(path.split("/")[1:]), name)
```
maxhutch: `os.path.join` is variadic, so I think you can do `os.path.join(dest_path, *(path.split("/")[1:]), name)`

citrination_client/client.py
```diff
@@ -238,6 +237,93 @@ def upload_file(self, file_path, data_set_id, root_path=None):
...
+        else:
+            return list_result
+
+    def move(self, dataset_id, original_path, target_path, is_dir=False):
+        return self.move_or_copy(dataset_id, "move", original_path, target_path, is_dir)
+
+    def copy(self, dataset_id, original_path, target_path, is_dir=False):
+        return self.move_or_copy(dataset_id, "copy", original_path, target_path, is_dir)
+
+    def move_or_copy(self, dataset_id, move_or_copy, original_path, target_path, is_dir=False):
```
maxhutch: Since `move` and `copy` are in the API, could this be "private" (e.g. `_move_or_copy`)

citrination_client/tests/test_client.py
```diff
@@ -1,31 +1,77 @@
...
-    inputs = [{"CHEMICAL_FORMULA": "AlCu"}, ]
-    resp = client.predict("betterdensitydemo", inputs)
-    prediction = resp['candidates'][0]['Density']
-    assert abs(prediction[0] - 5.786) < 0.1
+    def test_move_file(self):
+        path = self.test_file_root + "weird_extensions.woodle"
+        target_move_path = "TARGET_MOVE_PATH.txt"
+        self.client.upload_file(path, self.set_id)
+        self.client.move(self.set_id, path, target_move_path)
+        listed_files = self.client.list(self.set_id, target_move_path)["files"]
```
maxhutch: Also check that the file is no longer at `path`

citrination_client/client.py
mcgalcode: ahhhh good idea!

citrination_client/client.py
mcgalcode: Yes, it definitely should be.

citrination_client/client.py
```diff
@@ -191,43 +191,43 @@ def predict(self, model_name, candidates):
     def _get_predict_url(self, model_name):
         return self.api_url + '/csv_to_models/' + model_name + '/predict'
 
-    def upload_file(self, file_path, data_set_id, root_path=None):
+    def upload(self, dataset_id, source_path, dest_path=None):
```
sinistral: Is it at all a concern that this is an interface change to a public method?

citrination_client/client.py
```diff
@@ -191,43 +191,43 @@ def predict(self, model_name, candidates):
...
         """
-        Upload file to Citrination.
-
-        :param file_path: File path to upload.
-        :param data_set_id: The dataset id to upload the file to.
-        :return: Response object or return code if the file was not uploaded.
+        Upload a file, specifying source and dest paths a file (acts as the scp command)
+        :param source_path: The path to the file on the source host
+        :param dest_path: The path to the file where the contents of the upload will be written (on the dest host)
+        :return: The response object or a boolean False if the file was not uploaded
```
sinistral: Is it true that a boolean is returned?  It looks like a JSON message is returned in all cases.
(Out of interest, why do we return JSON?  All values are memory-local and not going over the wire, aren't they?)

citrination_client/client.py
```diff
@@ -191,43 +191,43 @@ def predict(self, model_name, candidates):
...
+                for name in files:
+                    path_without_root_dir = path.split("/")[1:] + [name]
+                    current_dest_path = os.path.join(dest_path, *path_without_root_dir)
+                    current_source_path = os.path.join(path, name)
+                    success = self.upload(dataset_id, current_source_path, current_dest_path)
+            message = {"message": "Upload of files is complete."}
+            return json.dumps(message)
+        elif os.path.isfile(source_path):
+            url = self._get_upload_url(dataset_id)
+            file_data = { "file": { "dest_path": str(dest_path), "src_path": str(source_path)}}
```
sinistral: The format of the post message has changed.  I assume that the service has been updated correspondingly, but does this mean that older versions of the client will no longer work with the service?

citrination_client/client.py
```diff
@@ -191,43 +191,43 @@ def predict(self, model_name, candidates):
...
-                with open(file_path, 'rb') as f:
+                with open(source_path, 'rb') as f:
                     r = requests.put(s3url, data=f)
                     if r.status_code == 200:
                         url_data = {'s3object': j['url']['path'], 's3bucket': j['bucket']}
                         requests.post(self._get_update_file_upload_url(j['file_id']),
                                       data=json.dumps(url_data), headers=self.headers)
                         message = {"message": "Upload is complete.",
-                                   "data_set_id": str(data_set_id),
+                                   "dataset_id": str(dataset_id),
```
sinistral: This is essentially an interface change in that the format of the return value has changed.  Is this not going to affect existing consumers?

citrination_client/client.py
```diff
@@ -238,6 +238,93 @@ def upload_file(self, file_path, data_set_id, root_path=None):
...
+        if isinstance(list_result["files"], list):
+            return len(list_result["files"])
+        else:
+            return list_result
+
+    def move(self, dataset_id, original_path, target_path, is_dir=False):
+        return self._move_or_copy(dataset_id, "move", original_path, target_path, is_dir)
+
+    def copy(self, dataset_id, original_path, target_path, is_dir=False):
+        return self._move_or_copy(dataset_id, "copy", original_path, target_path, is_dir)
```
sinistral: You might introduce a modicum of safety by using [enums](https://pypi.python.org/pypi/enum34) for the action, unless there's a desire to avoid pulling in additional dependencies.

citrination_client/client.py
```diff
@@ -238,6 +238,93 @@ def upload_file(self, file_path, data_set_id, root_path=None):
...
+        Move a file within the "filesystem" of a dataset on Citrination.
+        :param dataset_id: The ID of the dataset in which the move will occur
+        :param move_or_copy: A string equal to either "move" or "copy", indicating which action should be performed
+        :param original_path: The path (to a file or to a directory) to move
+        :param target_path: The location to move the file(s)
+        :param is_dir: Boolean indicating whether or not you are moving a directory
+        """
+        url = self.api_url + '/datasets/' + str(dataset_id) + '/' + move_or_copy
+        data = {
+            "move": {
```
sinistral: This makes for a confusing API.  Is there any reason why the message to the copy call shouldn't be called "copy"?

citrination_client/client.py
```diff
@@ -238,6 +238,93 @@ def upload_file(self, file_path, data_set_id, root_path=None):
...
+                "is_dir": is_dir,
+                "latest": latest
+            }
+        }
+        url = self.api_url + '/datasets/' + str(dataset_id) + '/download_files'
+        r = requests.post(url,data=json.dumps(data), headers=self.headers)
+        return json.loads(r.content.decode('utf-8'))
+
 
     def get_dataset_files(self, dataset_id, latest = False):
```
sinistral: It might help to change the name of the boolean to something like `lateset_version_only`.

citrination_client/client.py
mcgalcode: This looks like an interface change but the `upload` method is still defined below.

citrination_client/client.py
mcgalcode: The service has been updated, but I made some changes that make the service backward compatible with the old body.

citrination_client/client.py
mcgalcode: You are totally right. I will undo this.

citrination_client/client.py
sinistral: I completely missed the difference in the name; sorry, Max.

citrination_client/client.py
mcgalcode: You are right. There is no real need for this top level field anyway. I will remove it.

citrination_client/client.py
mcgalcode: Excellent idea.

citrination_client/client.py
mcgalcode: I don't have a good answer for you Marc. Maybe someone else can explain?

citrination_client/client.py
mcgalcode: Done!

mcgalcode:
@sinistral made some updates!

maxhutch:
@kjaym @jsomara could we either collapse this wavefunction or break out the unobjectionable pieces into a new PR?  There are some great tests in here and the upload to destination is a huge QoL improvement, neither of which are dependent on the rest of the file manipulation functionality.

kjaym:
@maxhutch I'm reviewing the other PR which I think is what you are referencing in your comment.

PR #77, opened 2017-03-24
Adding support for predictions on data views

maxhutch:
Maintains backward support by looking for 404 errors and trying the old csv2models-based URL.


maxhutch:
Thanks for supporting this on the site @mcgalcode !

PR #81, opened 2017-03-31
Adds Enhanced Upload and Listing Functionality to Client

mcgalcode:
NOTE: Since the fate of the 'move' and 'copy' commands is uncertain, I think it would be good to expose the other enhancements that that PR offers, separated from those. I have had some explicit requests for these features from data org. Let me know if this feature set has any ingestion implications that should prevent it from being implemented.

The intention was for this PR to only add upload with destination path specification and the list files ability.

  Adds the ability to specify the destination file name or folder
  that an upload will appear in in the dataset.

  Adds a command to list files in a dataset from the client based
  on an input glob.


citrination_client/client.py
```diff
@@ -224,43 +224,43 @@ def _get_predict_url(self, model_name):
...
         """
-        Upload file to Citrination.
-
-        :param file_path: File path to upload.
-        :param data_set_id: The dataset id to upload the file to.
-        :return: Response object or return code if the file was not uploaded.
+        Upload a file, specifying source and dest paths a file (acts as the scp command)
+        :param source_path: The path to the file on the source host
+        :param dest_path: The path to the file where the contents of the upload will be written (on the dest host)
+        :return: A JSON blob with a message indicating the status of the upload
```
kjaym: Can you add the fields that are returned and what they contain.

citrination_client/client.py
```diff
@@ -271,19 +271,78 @@ def upload_file(self, file_path, data_set_id, root_path=None):
                 message = {"message": "Upload failed.",
                                    "status": r.status_code}
                 return json.dumps(message)
+        else:
+            return json.dumps({
```
kjaym: The return objects are a little hard to parse. Can you add another field, something like `"success": boolean`, that can be easily parsed by someone to know whether the upload was successful.

citrination_client/client.py
```diff
@@ -271,19 +271,78 @@ def upload_file(self, file_path, data_set_id, root_path=None):
...
+    def upload_file(self, file_path, dataset_id, root_path=None):
+        """
+        Upload file to Citrination.
+        :param file_path: File path to upload.
+        :param data_set_id: The dataset id to upload the file to.
+        :return: Response object or return code if the file was not uploaded.
+        """
+        return self.upload(dataset_id, file_path)
+
+    def list(self, dataset_id, glob=".", is_dir=False):
```
kjaym: Is `glob` something that the average person understands? What about something like `match`?

citrination_client/client.py
```diff
@@ -271,19 +271,78 @@ def upload_file(self, file_path, data_set_id, root_path=None):
...
+        :param dataset_id: The ID of the dataset to search for files.
+        :param glob: A pattern which will be matched against files in the dataset.
+        :param is_dir: A boolean indicating whether or not the pattern should match against the beginning of paths in the dataset.
+        :return: Response object or return code if the file was not uploaded.
+        """
+        url = self.api_url + '/datasets/' + str(dataset_id) + '/list_filepaths'
+        data = {
+            "list": {
+                "glob": glob,
+                "is_dir": is_dir
```
kjaym: Convention elsewhere in the APIs is camel-case for JSON field names, snake-case for python variables.

citrination_client/client.py
```diff
@@ -271,19 +271,78 @@ def upload_file(self, file_path, data_set_id, root_path=None):
...
+
+    def list(self, dataset_id, glob=".", is_dir=False):
+        """
+        List matched filenames in a dataset on Citrination.
+        :param dataset_id: The ID of the dataset to search for files.
+        :param glob: A pattern which will be matched against files in the dataset.
+        :param is_dir: A boolean indicating whether or not the pattern should match against the beginning of paths in the dataset.
+        :return: Response object or return code if the file was not uploaded.
+        """
+        url = self.api_url + '/datasets/' + str(dataset_id) + '/list_filepaths'
```
kjaym: Related to point just below this, I don't know what the conventions are elsewhere in the URLs on citrination. I assume underscores are standard, but please make sure.

citrination_client/client.py
```diff
@@ -271,19 +271,78 @@ def upload_file(self, file_path, data_set_id, root_path=None):
...
+        data = {
+            "list": {
+                "glob": glob,
+                "is_dir": is_dir
+            }
+        }
+        r = requests.post(url, data=json.dumps(data), headers=self.headers)
+        return json.loads(r.content.decode('utf-8'))
+
+    def matched_file_count(self, dataset_id, glob=".", is_dir=False):
```
kjaym: Same comment about glob.

citrination_client/client.py
```diff
@@ -271,19 +271,78 @@ def upload_file(self, file_path, data_set_id, root_path=None):
...
+                "latest": latest_version_only
+            }
+        }
+        url = self.api_url + '/datasets/' + str(dataset_id) + '/download_files'
+        r = requests.post(url,data=json.dumps(data), headers=self.headers)
+        return json.loads(r.content.decode('utf-8'))
+
 
-    def get_dataset_files(self, dataset_id, latest = False):
+    def get_dataset_files(self, dataset_id, latest_version_only = False):
```
kjaym: This will break backwards compatibility, right?

mcgalcode: I don't think so. Isn't this variable only used in the scope of the method?

citrination_client/client.py
mcgalcode: The use of `datasets` and `data_sets` was not agreed upon for a long time so a while ago we decided to use `datasets` moving forward. The old routes which rely on data_sets still work but are marked as deprecated.

citrination_client/client.py
kjaym: I thinking more about scripts that would be calling this method _outside_ of this package. If those scripts are referring to `latest` then they would break, right?

citrination_client/client.py
kjaym: Oh, I was actually talking about list_filepaths

citrination_client/client.py
maxhutch: Does the average person use this API?  I think `glob` is a standard name for this.

citrination_client/client.py
mcgalcode: Oh, yes we use underscores everywhere in the routes in the ruby app.

citrination_client/client.py
mcgalcode: Ah you are totally right. I did not realize you could use named arguments at any time.

mcgalcode:
@kjaym made some changes

citrination_client/client.py
mcgalcode: I am going to leave this per MaxH, I think glob is a common way of referring to this

mcgalcode:
@kjaym ping

kjaym:
@mcgalcode @maxhutch There is a merge conflict here. Looks good otherwise.

kjaym:
@mcgalcode @maxhutch bump

mcgalcode:
@kjaym @maxhutch resolved merge conflict

PR #82, opened 2017-04-04
Feature/return csv

kjaym:
@maxhutch I'm adding in support for returning a CSV of results. This will basically create a CSV from the `extracted` fields on search hits.


kjaym:
One thing that we will need to do here is handle merging CSVs when we are auto-paging through results.

PR #85, opened 2017-04-12
predict for custom models

kjaym:
@maxhutch @sparadiso This just pushes into the master branch the ability to predict using a custom model. Erin is interested in getting this out. Any reason not to?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,5 @@
 language: python
 
-sudo: false
-
 env:
   global:
   - CITRINATION_API_KEY=IeReqZ4bK4TS1jjbg3q2zAtt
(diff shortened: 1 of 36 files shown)


sparadiso:
Erin has used this functionality to DOS the site...recently. Without rate limiting on the backend we should be aware that this is dangerous. That said, it is not any _more_ dangerous than what is already available to the public so I'm fine with this if it is urgent.

sinistral:
Adding a new DoS vector makes it _more_ dangerous  d:  At the _very_ least, given that we know that significant impact is possible, if not likely, we should have a good story around recovery and remediation.  Ideally we should protect ourselves, but at a minimum we should have an ops guide.

maxhutch:
Is @eantono still blocking on this?

eantono:
@maxhutch I would still like to use this functionality, what do you need from me on this? I thought it was blocked on rate limiting being implemented

PR #88, opened 2017-05-05
bugfix - id/classification query setters did not assign values

kjaym:
The setters in id_query and classification_query were not actually setting anything. This fixes that.

@eantono @maxhutch @mcgalcode


maxhutch:
LGTM

PR #89, opened 2017-06-07
Ml statuses

mcgalcode:


mcgalcode:
Note: I did not add tests for this because, without a way to add data views from the API, we would have to rely on some hardcoded data view ID references. While that is not implemented, my feeling is that we can use the citrination-api-tests to test these features

maxhutch:
:+1:

PR #92, opened 2017-06-14
Admin: test on production

maxhutch:
Moving the tests to production and re-enabling the prediction tests.  The API testing user's API token is encrypted.


.travis.yml
```diff
@@ -1,24 +1,19 @@
 language: python
-
 env:
   global:
-  - CITRINATION_API_KEY=IeReqZ4bK4TS1jjbg3q2zAtt
```
sinistral: Has this key been shredded?

citrination_client/tests/test_client.py
```diff
@@ -54,20 +54,41 @@ def test_upload_directory(self):
...
+  
+        resp = client.predict(vid, inputs)
+        prediction = resp['candidates'][0]
+        egap = '$\\varepsilon$$_{gap}$ ($\\varepsilon$$_{LUMO}$-$\\varepsilon$$_{HOMO}$)'
+        voltage = 'Open-circuit voltage (V$_{OC}$)'
+        assert 'Mass'  in prediction, "Mass prediction missing (check ML logic)"
+        assert egap    in prediction, "E_gap prediction missing (check ML logic)"
+        assert voltage in prediction, "V_OC prediction missing (check ML logic)"
+ 
+        assert abs(prediction['Mass'][0] - 250) < 50.0, "Mass mean prediction beyond tolerance (check ML logic)"
```
sinistral: What about wrapping tolerance logic in a function so that you can write something like:
```
assert within_tolerance(prediction['Mass'][0], +50.0, -0.0), ...
```
It makes the assertion a little more readable.

maxhutch:
@sinistral good call with the helper

.travis.yml
sinistral: https://citrine.myjetbrains.com/youtrack/issue/CTRN-71

PR #93, opened 2017-06-15
Feature: tsne

maxhutch:
Adding t-SNE call that hits the data analysis endpoint and then re-formats the keys a bit.

CC: @eantono thoughts?


citrination_client/client.py
```diff
@@ -304,6 +304,35 @@ def upload_file(self, file_path, dataset_id, root_path=None):
         """
         return self.upload(dataset_id, file_path)
 
+    def _data_analysis(self, model_name):
```
mcgalcode: I think nearly this exact method exists in `feature/dev-api` as a public method (and is used in the api tests). Should we keep this private or let users consume data analyses directly?

citrination_client/client.py
```diff
@@ -516,3 +545,10 @@ def _get_create_data_set_version_url(self, data_set_id):
         :return: URL for creating new data set versions.
         """
         return self.api_url+'/data_sets/'+str(data_set_id)+'/create_dataset_version'
+
+
+    def _get_data_analysis_url(self, model_name):
+        return self.api_url + '/data_views/' + model_name + '/data_analysis'
+
+    def _get_estimators_url(self, model_name):
+        return self.api_url + '/data_views/' + model_name + '/estimators'
```
mcgalcode: Is this related to the other content in this PR somehow?

maxhutch: No; this bled through.  Removing.

citrination_client/client.py
maxhutch: I don't think the raw output from the dev-api is user friendly, which is why I "hid" it.  In a future PR, we can expose the correlations as another method.

eantono:
The dictionary format looks reasonable to me :+1:

citrination_client/client.py
mcgalcode: Sounds good to me

citrination_client/client.py
mcgalcode: 👍

PR #98, opened 2017-07-21
Add Necessary Headers To Upload PUT Requests to S3

mcgalcode:
By requiring the use of server side encryption in the buckets used in the
  back end, we require that the `x-amz-acl` and `x-amz-server-side-encryption`
  headers be present on the request to the presigned url that uploads files.

  This commit adds the necessary headers.

### DIFF
--- a/citrination_client/client.py
+++ b/citrination_client/client.py
@@ -269,7 +269,7 @@ class CitrinationClient(object):
                 j = json.loads(r.content.decode('utf-8'))
                 s3url = self._get_s3_presigned_url(j)
                 with open(source_path, 'rb') as f:
-                    r = requests.put(s3url, data=f)
+                    r = requests.put(s3url, data=f, headers=self._upload_put_url_headers())
                     if r.status_code == 200:
                         url_data = {'s3object': j['url']['path'], 's3bucket': j['bucket']}
                         requests.post(self._get_update_file_upload_url(j['file_id']),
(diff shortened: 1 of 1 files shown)


citrination_client/client.py
```diff
@@ -304,6 +304,18 @@ def upload_file(self, file_path, dataset_id, root_path=None):
...
 
+    def _upload_put_url_headers(self):
+        """
+        Returns headers necessary for requests to presigned put urls
+        :return: A dictionary representing the headers required for requests
+                to presigned URLs for uploading files
+        """
+        headers = {
+            "x-amz-server-side-encryption": "AES256",
+            "x-amz-acl": "private"
```
sinistral: Is this _strictly necessary_?  I'm not objecting to it's presence, just wondering if it is _necessary_ for the uploads to work.

jsomara: My understanding is that the default acl is private so it is not necessary

mcgalcode: It is necessary because the way we are generating the signature requires this.

mcgalcode: (I have tested this and you need both headers in order for the request to succeed)

sinistral: > because the way we are generating the signature requires this

What are we doing specially that requires this?

mcgalcode: When we generate the presigned url we pass `acl: "private"` as an option to it.

PR #99, opened 2017-07-21
Add Necessary Headers To Upload PUT Requests to S3

mcgalcode:
By requiring the use of server side encryption in the buckets used in the
  back end, we require that the `x-amz-acl` and `x-amz-server-side-encryption`
  headers be present on the request to the presigned url that uploads files.

  This commit adds the necessary headers.

### DIFF
--- a/citrination_client/client.py
+++ b/citrination_client/client.py
@@ -60,7 +60,7 @@ class CitrinationClient(object):
                     hits.extend(partial_results.hits)
             return PifSearchResult(hits=hits, total_num_hits=total, took=time)
 
-        response = requests.post(self.pif_search_url, data=pif.dumps(pif_query), headers=self.headers)
+        response = self._post_with_version_check(self.pif_search_url, data=pif.dumps(pif_query), headers=self.headers)
         if response.status_code != requests.codes.ok:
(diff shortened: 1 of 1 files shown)


jsomara:
Confirmed working against backported Ruby AWS SDK 2.0.39 

cc @sparadiso

jsomara:
`python-requests` puts the following request in the outbound header:

`Agent': 'python-requests/2.14.2'`

citrination_client/client.py
```diff
@@ -460,6 +448,30 @@ def _get_upload_url(self, data_set_id):
...
+        result = requests.put(s3url, data=data, headers=headers)
+        return self._check_response_for_version_mismatch(result)
+
+    def _check_response_for_version_mismatch(self, response):
+        response_content = json.loads(response.content.decode('utf-8'))
+        try:
+            if response.status_code == 400:
+                error_type = response_content["error_type"]
+                if error_type == "Version Mismatch":
+                    print("Version mismatch with Citrination identified. Please check for available PyCC updates")
```
kjaym: Rather than hard code the message in the library, can we have a peer to "error_type" in the response that is something like "error_message"? That gives us more control over signaling to the user what they need to do.

PR #107, opened 2017-08-16
Bugfix with creating new dataset version

kjaym:
The wrong number of arguments were being passed into the post method.


maxhutch:
FYI: This needs to be merged into `develop` as well.

PR #117, opened 2017-09-15
Support for updated query syntax.

kjaym:
This adds support for the new query syntax that is used on Citrination. This adds a PifSystemReturningQuery and DatasetReturningQuery object that share a common parent. This allows for a common interface to querying for PIFs or datasets by the content of either/both. The only difference between these two returning queries is the formatting that they allow that is specific to the type that each returns.

The set of features in this version matches exactly what is available in JCC 2.4.0, the current release version of that package.

All deployments of Citrination currently support the updated syntax. Servers are backwards compatible with v1.x and v2.x of PyCC, but v3.x adds significantly improved control on queries.


kjaym:
Rebuild me!

setup.py
```diff
@@ -1,7 +1,7 @@
 from setuptools import setup, find_packages
 
 setup(name='citrination-client',
-      version='2.1.0',
+      version='3.0.0',
```
maxhutch: Save this for the release branch

citrination_client/search/pif/result/pif_search_hit.py
```diff
@@ -82,6 +89,18 @@ def score(self):
         self._score = None
 
     @property
+    def updated_at(self):
```
maxhutch: Could you add a test for this property?

citrination_client/search/pif/result/pif_search_hit.py
```diff
@@ -113,3 +132,15 @@ def extracted(self, extracted):
     @extracted.deleter
     def extracted(self):
         self._extracted = None
+
+    @property
+    def extracted_path(self):
```
maxhutch: Could you add a test for this property?

citrination_client/search/pif/query/pif_system_query.py
```diff
@@ -86,6 +96,18 @@ def uid(self):
         self._uid = None
 
     @property
+    def updated_at(self):
```
maxhutch: Could you add a test for this property?

citrination_client/search/core/query/data_query.py
```diff
@@ -0,0 +1,91 @@
...
+    """
+    Query against dataset metadata, PIF content, file content, or some combination of those types.
+    """
+
+    def __init__(self, logic=None, simple=None, dataset=None, system=None, query=None, **kwargs):
+        """
+        Constructor.
+
+        :param logic: The logic to apply to the query ('SHOULD', 'MUST', 'MUST_NOT', or 'OPTIONAL').
+        :param simple: String with the simple search to run against all fields.
```
maxhutch: Could you add a test for simple queries?

maxhutch: bump

citrination_client/search/pif/query/pif_system_query.py
maxhutch: bump

citrination_client/search/pif/result/pif_search_hit.py
maxhutch: bump

citrination_client/search/pif/result/pif_search_hit.py
maxhutch: bump

citrination_client/search/core/query/data_query.py
kjaym: @maxhutch added

citrination_client/search/pif/query/pif_system_query.py
kjaym: @maxhutch added

citrination_client/search/pif/result/pif_search_hit.py
kjaym: @maxhutch added

citrination_client/search/pif/result/pif_search_hit.py
kjaym: @maxhutch added

PR #122, opened 2017-12-14
Release/3.1

kjaym:


kjaym:
@maxhutch This is everything that was just merged into develop with a version bump added.

Issue #75, opened 2017-03-14
Utility to paginate through results for users


sparadiso:
We should provide users with a utility that paginates for them - pagination is an engineering construct, not a meaningful materials science concept. As a materials scientist, it surprises me that I can't ask for information and retrieve the list of results immediately. I'd rather provide a PifQuery and a total record count and get back a list of hits equal to (or less than) the requested record count.

kjaym:
If you don't ask for a max number of records or set a starting index then it will iterate for you.

sparadiso:
```
resp = client.search(PifQuery(include_datasets=[150146], size=1500, random_results=True, return_system=True))

lst = list()
for hit in resp.hits:
    lst.append(hit.system.uid)
    print len(lst)
```

That code prints up to 500.

sparadiso:
Put another way, I think that's an example of an intuitive request with an unintuitive response (although I understand why it behaves this way).

kjaym:
Sure, but if you don't put a size, then it does behave the way that you are suggesting. You could update https://github.com/CitrineInformatics/python-citrination-client/blob/develop/citrination_client/client.py#L47-L66 If you want to only check for None-valued from_index.

sparadiso:
That still leaves my request unsupported. IMO the following code should work as expected:

```
resp = client.search(PifQuery(include_datasets=[150146], size=1500, return_system=True))

lst = list()
for hit in resp.hits:
    lst.append(hit.system.uid)
print len(lst)

>> 1500
```

sparadiso:
n=1 :p

sparadiso:
(note the motivation here is that we're starting to actually get large datasets on the site, so getting a representative subsample is more likely to be a common request)

kjaym:
If you look at line 47 of the above, remove `pif_query.size is None`, then slightly adjust the logic on line 49 to take into account the requested size, then you'll get exactly that behavior.

maxhutch:
@sparadiso I can handle this if you're not working on it.  Just assign it to me.

sparadiso:
_we have the technology_

sparadiso:
I will not have time to do it until EOD at earliest (I'm done w/ pcc too, though). Feel free!

jling11:
This whole chain makes me both laugh and cry.  I want this fix to happen.

jling11:
Is there a jira card for this?

sparadiso:
I don't think so.

PR #125, opened 2018-01-11
Paginate even when from and size are set

maxhutch:
@kjaym how do you feel about this?  If its :+1:, I'll go ahead and write tests.

This addresses #75,  @jling11


kjaym:
@maxhutch Seems fine. Can you add the same logic to dataset search.

mcgalcode:
closing since this is wrapped up in the refactor PR

PR #126, opened 2018-01-24
Allow proxies to be supplied to the client

kjaym:
In case users have to proxy through a URL, this adds support for including the proxy list in a CitrinationClient instance.

### DIFF
--- a/citrination_client/client.py
+++ b/citrination_client/client.py
@@ -17,7 +17,7 @@ class CitrinationClient(object):
     Class for interacting with the Citrination api.
     """
 
-    def __init__(self, api_key, site='https://citrination.com'):
+    def __init__(self, api_key, site='https://citrination.com', proxies=None):
         """
         Constructor.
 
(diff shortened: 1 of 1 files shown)


citrination_client/client.py
```diff
@@ -17,20 +17,23 @@ class CitrinationClient(object):
     Class for interacting with the Citrination api.
     """
 
-    def __init__(self, api_key, site='https://citrination.com'):
+    def __init__(self, api_key, site='https://citrination.com', proxies=None):
```
sinistral: Why not have the default be an empty map rather than `None`?  It'll save having to do the explicit check later on.

kjaym: I would want to keep the check in case a user passed in `proxies=None`

sinistral: Does that actually have consequences?  Does it `requests` to pass `None` rather than empty map?

kjaym: It turns out that both `{}` and `None` work just fine with requests. I just removed the if/else check and just pass through the user input, defaulting to None.

kjaym:
@sinistral We have tested configuring requests directly with a proxy.

PR #127, opened 2018-01-24
Release/3.2.0

kjaym:
This includes support for supplying proxies to CitrinationClient.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,7 @@
 from setuptools import setup, find_packages
 
 setup(name='citrination-client',
-      version='3.1.0',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -1,7 +1,7 @@
 from setuptools import setup, find_packages
 
 setup(name='citrination-client',
-      version='3.1.0',
+      version='3.2.0',
```
sinistral: Good catch; I didn't think about that!

kjaym: This is our usual pattern. Merge into develop without version changes then create a release branch that includes the version bump.

PR #134, opened 2018-03-22
Catch Timeouts In Queries

mcgalcode:


maxhutch:
@mcgalcode did you mean for this to be a hotpatch?

PR #137, opened 2018-03-23
Support 204 return for query timeout as well as 524

kjaym:
We are going to start returning a 204 to indicate a query timeout rather than a 524.

### DIFF
--- a/citrination_client/client.py
+++ b/citrination_client/client.py
@@ -69,7 +69,7 @@ class CitrinationClient(object):
                     hits.extend(partial_results.hits)
             return PifSearchResult(hits=hits, total_num_hits=total, took=time)
 
-        response = self._post_with_version_check(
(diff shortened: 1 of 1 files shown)


mcgalcode:
204 might not indicate timeout for every request. For instance, 204 is also the correct code for the response from deleting a resource. Is there an easy way to move that check to the search method? If not, I can wrap this up in the new PyCC since we don't have any other 204 endpoints for now.

kjaym:
Good point ... incoming change

kjaym:
@mcgalcode This adds new methods for dealing with search requests that know that a 204 means a query timeout. This will be a little more contained in the next breaking version that keeps a specific search client.

mcgalcode:
lgtm, will port to 4.x pr

PR #141, opened 2018-04-11
Feature/add url field

sparadiso:
Add url field to search.


kjaym:
@sparadiso Please make sure that @mcgalcode knows about this and that it needs to go into v4 of pycc

sparadiso:
We might actually want to card this up and follow through on it just for the rep adding an indexed field to the PIF.

Thoughts @andrewmillspaugh @mcgalcode @sinistral ?

mcgalcode:
@sparadiso We should definitely follow through on this. Is there more work to be done in other repositories before this is supported?

PR #145, opened 2018-05-04
Feature/doc updates

joannehill:
Addresses:

(https://citrine.atlassian.net/browse/CITR-1881) Add 50k limit to docs and raise an exception when exceeded

and

(https://citrine.atlassian.net/browse/CITR-1882) Add to doc what .get_value returns if the key isn’t present for a prediction

and

(https://citrine.atlassian.net/browse/CITR-1886) Removing mention of get_matched_dataset_files


citrination_client/search/client.py
```diff
@@ -39,6 +39,10 @@ def pif_search(self, pif_system_returning_query):
         :return: :class:`PifSearchResult` object with the results of the query.
         :rtype: :class:`PifSearchResult`
         """
+
+        if sum(filter(None, [pif_system_returning_query.from_index, pif_system_returning_query.size])) >= 50000:
```
andrewmillspaugh: What is the purpose of the filter here? Maybe I'm missing something, but couldn't you just say:
```
if pif_system_returning_query.from_index + pif_system_returning_query.size >= 50000
```
Also, I would define the 50000 in a constant at the top of the files. MAX_QUERY_RESULTS or something like that.

Finally, I don't actually understand why from_index is being added to size. Is it because we don't allow access past 50k entries? If so, I would probably format the error message such that that is more clear.

citrination_client/search/dataset/query/dataset_returning_query.py
```diff
@@ -3,7 +3,7 @@
 
 class DatasetReturningQuery(BaseReturningQuery):
     """
-    Query used to return information about datasets.
+    Query used to return information about datasets. Results limited to 50,000. Please chose values for from_index and size that do not exceed this limit.
```
andrewmillspaugh: Don't include the 50k number in here. That's asking to become out of date.

citrination_client/search/dataset/tests/test_dataset_query.py
```diff
@@ -38,6 +38,14 @@ def test_dataset_search(self):
                     id=Filter(equal='151278')))))
         assert 1 == response.total_num_hits
 
+    def test_search_limit_enforced_dataset_search(self):
```
andrewmillspaugh: Here, if you had the MAX_QUERY_RESULTS defined as a constant, you could just import that instead of using a hard coded value here.

citrination_client/search/client.py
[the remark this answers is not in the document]
sparadiso: I believe 50k is the default for all deployments. However, it's probably not a bad thing to expose through a lightweight API route in CSL that can be proxied through the FE.

sparadiso: I would guess that this is to handle the case where `from_index` or `size` are null or None?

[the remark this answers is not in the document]
andrewmillspaugh: https://en.m.wikipedia.org/wiki/Zen_of_Python

andrewmillspaugh: https://en.m.wikipedia.org/wiki/Zen_of_Python

> Explicit is better than implicit; If the implementation is hard to explain, it's a bad idea.

mcgalcode: @joannehill I assume sean is right about your intent here; maybe add a comment explaining?

citrination_client/search/dataset/query/dataset_returning_query.py
mcgalcode: In principle, I agree with you. However, this is a common issue for users and we had a specific request that it be documented. Once we add an API endpt that can return this value, we can reconsider but for now I would weight more heavily communicating correctly over avoiding risk of obsolescence.

citrination_client/search/dataset/tests/test_dataset_query.py
mcgalcode: @joannehill seconded.

citrination_client/search/client.py
```diff
@@ -13,6 +14,7 @@
 import requests
 
 DEFAULT_FAILURE_MESSAGE="An error occurred requesting search results from Citrination"
+MAX_QUERY_RESULTS=50000
```
andrewmillspaugh: Add spaces around the `=`

citrination_client/search/client.py
```diff
@@ -42,8 +42,9 @@ def pif_search(self, pif_system_returning_query):
         :rtype: :class:`PifSearchResult`
         """
 
-        #filter to get the values for from_index and size that are not None and add them to check that the max index supported will not be exceeded
-        if sum(filter(None, [pif_system_returning_query.from_index, pif_system_returning_query.size])) >= MAX_QUERY_RESULTS:
+        start_index = pif_system_returning_query.from_index or 0
+        size = pif_system_returning_query.size or 0
+        if start_index + size > MAX_QUERY_RESULTS:
```
andrewmillspaugh: <3

citrination_client/search/client.py
```diff
@@ -39,6 +41,12 @@ def pif_search(self, pif_system_returning_query):
         :return: :class:`PifSearchResult` object with the results of the query.
         :rtype: :class:`PifSearchResult`
         """
+
+        start_index = pif_system_returning_query.from_index or 0
+        size = pif_system_returning_query.size or 0
+        if start_index + size > MAX_QUERY_RESULTS:
+            raise CitrinationClientError("Citrination does not support pagination past the {}th result. Please change from_index and size arguments to be within this limit".format(MAX_QUERY_RESULTS))
```
mcgalcode: This validation happens the same way in two methods; I recommend that you extract it into a helper method so you don't have to repeat yourself.

citrination_client/search/client.py
```diff
@@ -32,6 +32,19 @@ def _handle_response(self, response, failure_message=DEFAULT_FAILURE_MESSAGE):
 
         return super(SearchClient, self)._handle_response(response, failure_message)
 
+    def _validate_query_depth(self, pif_system_returning_query):
```
andrewmillspaugh: This is totally fine. However, for the sake of learning about some other ways you could improve it...
- You could change the name to just `_validate_query`, and then use that for any validation that you might want to add in the future.
- You could rename either of `_validate_query` or `_execute_paginating_search` such that they have more consistent names. I would probably choose something like `_validate_search_query` and `_execute_search_query`, but the important thing is that they describe reality and are consistent, for the sake of the reader.
- You could consider moving the call to the `_validate_query` method to inside of the `_execute_query` method, so that any query is automatically validated, as opposed to having to call those things separately. There are reasonable concerns with that approach, though (namely, what if I wanted to execute a non validated query?).
- You could improve the wording of the error message to be more explicit, E.g. `Reduce either of from_index or size arguments such that their sum is below {MAX_QUERY_DEPTH}`

citrination_client/search/pif/tests/test_pif_query.py
```diff
@@ -47,7 +47,7 @@ def test_search_limit_enforced_pif_search(self):
         """
         Tests that if a user tries to access more than the max allowed results an error is thrown
         """
-        query = PifSystemReturningQuery(from_index=MAX_QUERY_RESULTS,size=10)
+        query = PifSystemReturningQuery(from_index=MAX_QUERY_DEPTH,size=10)
```
andrewmillspaugh: Space after the comma

citrination_client/search/client.py
```diff
@@ -30,6 +32,19 @@ def _handle_response(self, response, failure_message=DEFAULT_FAILURE_MESSAGE):
...
+        """
+        Checks to see that the query will not exceed the max query depth
+
+        :param pif_system_returning_query: The PIF system query to execute.
+        :type pif_system_returning_query: :class:`PifSystemReturningQuery`
+        """
+
+        start_index = pif_system_returning_query.from_index or 0
+        size = pif_system_returning_query.size or 0
+        if start_index + size > MAX_QUERY_DEPTH:
```
andrewmillspaugh: You could also a check here to make sure start_index and size are non-negative integers, though maybe they've already been validated somewhere else.

citrination_client/search/client.py
```diff
@@ -123,8 +123,8 @@ def _search_internal(self, returning_query, result_class):
             failure_message = "Error while making dataset search request"
 
         response_json = self._get_success_json(self._post(
-                    route, data=json.dumps(returning_query, cls=QueryEncoder),
-                failure_message=failure_message))
+                    route, data = json.dumps(returning_query, cls = QueryEncoder),
```
andrewmillspaugh: I *believe* [PEP8](https://www.python.org/dev/peps/pep-0008/) (the de facto Python style guide) says that keyword args like this do not have spaces around the equals sign. Personally, I don't actually care that much. But just FYI.

citrination_client/search/client.py
```diff
@@ -136,13 +136,13 @@ def pif_multi_search(self, multi_query):
...
         failure_message = "Error while making PIF multi search request"
-        response_dict = self._get_success_json(self._post(routes.pif_multi_search, data=json.dumps(multi_query, cls=QueryEncoder), failure_message=failure_message))
+        response_dict = self._get_success_json(self._post(routes.pif_multi_search, data = json.dumps(multi_query, cls = QueryEncoder), failure_message = failure_message))
 
         return PifMultiSearchResult(**keys_to_snake_case(response_dict['results']))
 
-    def generate_simple_chemical_query(self, name=None, chemical_formula=None, property_name=None, property_value=None,
-                               property_min=None, property_max=None, property_units=None, reference_doi=None,
-                               include_datasets=[], exclude_datasets=[], from_index=None, size=None):
+    def generate_simple_chemical_query(self, name = None, chemical_formula = None, property_name = None, property_value = None,
```
andrewmillspaugh: I believe PEP8 also has strong feelings about multiline function signatures. This one might be worth digging into.

citrination_client/search/client.py
```diff
@@ -48,7 +48,7 @@ def _validate_search_query(self, pif_system_returning_query):
         if size < 0:
             raise CitrinationClientError("Size cannot be negative. Please enter a value greater than or equal to zero")
         if start_index + size > MAX_QUERY_DEPTH:
-            raise CitrinationClientError("Citrination does not support pagination past the {}th result. Please change from_index and size arguments to be within this limit".format(MAX_QUERY_DEPTH))
+            raise CitrinationClientError("Citrination does not support pagination past the {}th result. Please reduce either the from_index and/or size such that their sum is below {}".format(MAX_QUERY_DEPTH, MAX_QUERY_DEPTH))
```
andrewmillspaugh: I believe you can just use MAX_QUERY_DEPTH once if you write your format string like:
`"blah blah blah {0} blah blah blah {0}".format(variable)`

andrewmillspaugh: And, if we didn't have to support old python, you could do: `f"blah blah blah {variable} blah blah blah {variable}"`

citrination_client/search/client.py
joannehill: Thanks, I will switch back to PyCharm

andrewmillspaugh:
<3

PR #146, opened 2018-05-05
Adding classes from the internal sdk without modification

joannehill:
In support of CITR-1915 we are adding these classes so that we can provide users access to the data view summary information about each column. Actual support for getting this summary information will be done in a separate PR.


maxhutch:
@joannehill the import paths to refer to the internal SDK.

maxhutch:
The python2.7 test is passing but 3.x are failing.

citrination_client/models/columns/__init__.py
```diff
@@ -0,0 +1,8 @@
+from base import BaseColumn
```
mcgalcode: These import statements will cause tests to fail in python 3.x.

I think you need to either change them to use the full path (e.g. `from citrination_client.models.columns.base import BaseColumn`), or use the relative module path syntax (e.g. `from .base import BaseColumn`)

joannehill: Thanks, yup. Will replace with full paths

mcgalcode:
@hutch any other comments here?

PR #148, opened 2018-05-07
Feature/model status

joannehill:
CITR-1916 (PyCC get model status endpoint wrapper)

`test_data_view_summary` will fail until the endpoint is available on public Citrination.

Retrieves the status for all of the services associated with a data view:
            - predict
            - experimental_design
            - data_reports
            - model_reports


citrination_client/models/tests/test_event.py
```diff
@@ -0,0 +1,17 @@
...
+
+    e = Event(
+        title = title,
+        subtitle = subtitle,
+        normalized_progress = progress
+    )
+
+    assert e.title == title
+    assert e.subtitle == subtitle
+    assert e.normalized_progress == progress
```
maxhutch: Test `is_ready` in multiple states.

citrination_client/models/service_status.py
```diff
@@ -0,0 +1,115 @@
...
+        :type reason: str
+        :param event: An event object describing the current state of the service's
+            progress toward readiness
+        :type event: Event
+        """
+        self._ready   = ready
+        self._context = context
+        self._event   = event
+        self._reason  = reason
+
```
maxhutch: Can we validate the progress of the event against the `ready` value?

citrination_client/models/tests/test_models_client.py
```diff
@@ -174,4 +174,22 @@ def test_can_submit_run_with_no_target():
...
+@pytest.mark.skipif(environ['CITRINATION_SITE'] != "https://citrination.com", reason="Data view status tests only supported on open")
+def test_data_view_status_reports_services_ready():
+    """
+    Tests that a data view on open citrination which has successfully trained reports
+    the status of it's services as ready.
+    """
+
+    view_id = "524"
+
+    status = client.get_data_view_service_status(data_view_id=view_id)
```
maxhutch: Can we add this call and the following asserts to a few other of the tests (like retrain) to make sure it works for a non-ready status?

citrination_client/models/service_status.py
mcgalcode: Can you elaborate on this hutch? We just accept and show whatever CMC has most recently pushed (which has historically included nonsensical combinations of progress + ready).

citrination_client/models/tests/test_models_client.py
mcgalcode: @maxhutch `retrain` is in the internal SDK, but not in PyCC. Unfortunately PyCC has no functionality yet for triggering model retrains (other than adding more data to an underlying data view). We could find a permanently broken view though.

citrination_client/models/service_status.py
maxhutch: Could we at least warn on a nonsensical combination of progress + ready so we can start to feed that signal back into the status produces?

citrination_client/models/tests/test_models_client.py
maxhutch: Then let's just make a note to use this in the internal SDK once its merged.

mcgalcode: Sounds good.

citrination_client/models/service_status.py
mcgalcode: That makes sense to me. @joannehill this would look like warning whenever `ready` being `True` anytime that progress is present and less than 1.0.

joannehill: @maxhutch would that just involve checking that ready is only `True` when the normalized_progress is `1`?

joannehill: :+1:

citrination_client/models/service_status.py
```diff
@@ -0,0 +1,125 @@
...
+        return ServiceStatus(
+            ready = response_dict["ready"],
+            reason = response_dict["reason"],
+            context = response_dict["context"],
+            event = event
+        )
+
+    @staticmethod
+    def _validate_response_dict(response_dict):
```
maxhutch: Why only apply this when building `from_response_dict`?  A future developer is free to create one of these objects with `__init__`, and might be confused when they don't get member validation.

mcgalcode:
LGTM

PR #149, opened 2018-05-07
Adds all_responses To PredictionResponse

mcgalcode:
Adds a method to list all the values stored inside a
  prediction result so there is no guessing involved.


citrination_client/models/prediction_result.py
```diff
@@ -31,4 +31,15 @@ def get_value(self, key):
         try:
             return self._values[key]
         except KeyError:
-            return None
\ No newline at end of file
+            return None
+
+    def all_responses(self):
```
maxhutch: Can we call it something more generic, like `all_names` or `all_keys`?

mcgalcode: Yes!

PR #151, opened 2018-05-09
Feature/download files

joannehill:
CITR-1887 (Download file and download all files - preserving directory hierarchy)


citrination_client/data/client.py
```diff
@@ -186,6 +186,32 @@ def get_dataset_file(self, dataset_id, file_path, version = None):
...
+        :type destination: str
+        :param chunk: Whether or not to chunk the file. Default True
+        :type chunk: bool
+        """
+        if not isinstance(dataset_files, list):
+            dataset_files = [dataset_files]
+
+        for f in dataset_files:
+            filename = f.path
+            local_path = '/'.join([destination, filename])
```
mcgalcode: You might consider using os.join here (then we'd handle windows users a bit better as well). There is a situation where `filename` _starts_ with `/` since users can specify any upload path. This should probably be resolved in the API, but you could handle it here by conditionally trimming leading slashes from filename.

PR #153, opened 2018-05-09
Fixes And Adds Tests For Simple Query Generation

mcgalcode:
Adds a method that got removed a while ago that had
  broken an untested method. Also adds tests for it the previously broken method.


mcgalcode:
@maxhutch that seems like a good idea, do you mean in two separate PRs?

PR #161, opened 2018-05-13
Use a valid dataset in tests

Setogit:
There are 4 test failures in the release 4.1.0.  This commit uses dataset 150675 and Chemical formula: AlCu.  With this commit, all the tests pass.


Setogit:
Travis will most likely fail because travis does not allow external PRs to use the encrypted CITRINATION_API_KEY.  I locally ran purest using my api key and confirmed it passed.

PR #162, opened 2018-05-13
Resync develop

joannehill:


maxhutch:
This should be from the release branch, as in #163

PR #164, opened 2018-05-14
Fix Failing Tests

mcgalcode:
- ensure that simple query generation test does not run
    unless the tests are being run against public citrination

  - give update dataset metadata test a 5 minute timeout window
    since we were experiencing many timeout failures before


mcgalcode:
Oh the irony.

PR #181, opened 2018-10-01
Adding method to get the latest template version

joannehill:


citrination_client/models/tests/test_models_client.py
```diff
@@ -68,6 +68,15 @@ def test_predict():
...
     _assert_prediction_values(prediction_result)
 
+@pytest.mark.skipif(environ['CITRINATION_SITE'] != "https://citrination.com", reason="Predict tests only supported on open citrination")
+def test_template_latest_version():
+    """
+    Tests that the latest version of the template can be returned
+    """
+    vid = 177
+    latest_template = client.template_latest_version('view_ml_{}_1'.format(vid))
+    assert latest_template == 1635
```
sparadiso: Isn't this going to fail immediately after that view gets retrained through the API? Or does the version not get bumped anymore?

joannehill: That is a good question. I didn't even think about the fact that would increment. I will just make this check that an integer is returned.

sparadiso: Either that or create a new view, make sure the version is `1`, and delete it.

sparadiso: If you go the latter route, then we should define a fixture so that any time you need a new vanilla view to test the API against you use the same one and only create it once per api test run.

PR #190, opened 2018-10-24
Use os.path.relpath To Determine Directory Nesting

mcgalcode:
By trying to write our own path splitting logic, we were running
  into problems with duplicate directory nesting in paths. By using
  relpath we delegate that logic to a standard library.

  Specifically, this solves the case where you try to upload a directory
  without adding a trailing slash to the path. Now it doesn't matter
  if you leave it out or add it in.


citrination_client/data/tests/test_data_client.py
```diff
@@ -21,8 +21,8 @@ def random_string():
 # one dataset with the same name
 dataset_name = "Tutorial dataset " + random_string()
 dataset_id = client.create_dataset(name=dataset_name, description="Dataset for tutorial").id
-test_file_root = './citrination_client/data/tests/test_files/'
-test_file_data_root = './citrination_client/data/tests/test_files/data/'
+test_file_root = 'citrination_client/data/tests/test_files/'
```
maxhutch: Why'd you remove these `./`?  Could you leave one in and one out for coverage?

citrination_client/data/tests/test_data_client.py
```diff
@@ -59,6 +59,15 @@ def test_upload_pif():
...
         assert json.loads(fp.read())["uid"] == pif.uid
 
+def test_does_not_require_trailing_slash():
+    src_path = "{}data_holder".format(test_file_data_root)
+    result = client.upload(dataset_id, src_path)
+    assert result.successful()
+
+    files = client.list_files(dataset_id)
+    first_path = files[0]
+    assert "data_holder/data_holder" not in first_path
```
maxhutch: The `files[0]` followed by a default-negative check seems easy to pass by accident.  What's to say this is the first file in the dataset?

mcgalcode: If any of them have that pattern, they will all have that pattern.

mcgalcode: (this is part of the path, not the file name)

citrination_client/data/tests/test_data_client.py
mcgalcode: Yep.

mcgalcode:
@maxhutch updated

PR #194, opened 2018-11-07
Add proxies param to the CitrinationClient (base)

sparadiso:
This PR adds an optional `proxies` parameter to the CitrinationClient. The proxies dict will be passed along to the requests module for all HTTP requests.

### DIFF
--- a/citrination_client/base/base_client.py
+++ b/citrination_client/base/base_client.py
@@ -15,7 +15,7 @@ class BaseClient(object):
     Base class that holds the universal constructor, utilities, etc
     """
 
-    def __init__(self, api_key, webserver_host, api_members=[], suppress_warnings=False):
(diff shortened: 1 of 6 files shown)


citrination_client/base/base_client.py
```diff
@@ -28,10 +28,18 @@ def __init__(self, api_key, webserver_host, api_members=[], suppress_warnings=Fa
         :param suppress_warnings: A flag indicating whether or not warning
             messages to stdout should be printed
         :type suppress_warnings: bool
+        :param proxies: proxies to use when making HTTP requests. E.g.,
+                proxies = {
+                  'http': 'http://10.10.1.10:3128',
```
genfx999: Where do these IP:PORT numbers come from?

sparadiso: Hah good catch. They come from the requests API doc (http://docs.python-requests.org/en/latest/user/advanced/#proxies)

PR #206, opened 2018-11-10
Update docs => gh-pages

sparadiso:
I think I did this right? The intent here was to update the documentation. I went into the develop branch, ran `make html` in the docs directory, then copied the results into the gh-pages branch root.


mcgalcode:
That sounds right to me.

PR #211, opened 2018-12-19
Add target response value support

sparadiso:
This just allows users to communicate specific value constraints to us through the API

### DIFF
--- a/citrination_client/models/design/tests/test_target.py
+++ b/citrination_client/models/design/tests/test_target.py
@@ -16,3 +16,4 @@ def test_target_initialization():
     # These initializations should not throw an error
     Target(name="Band gap", objective="Min")
     Target(name="Band gap", objective="Max")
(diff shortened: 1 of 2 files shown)


citrination_client/models/design/tests/test_target.py
```diff
@@ -15,4 +15,5 @@ def test_target_initialization():
 
     # These initializations should not throw an error
     Target(name="Band gap", objective="Min")
-    Target(name="Band gap", objective="Max")
\ No newline at end of file
+    Target(name="Band gap", objective="Max")
```
mcgalcode: Looks like we changed the cases on Min and Max - is that an API change that would affect existing scripts users have written?

sparadiso: No - the backend already converts to lower.

sparadiso: Good catch, though!

PR #214, opened 2019-01-09
CT-180 Feature/data views api

genfx999:


citrination_client/views/client.py
```diff
@@ -12,3 +14,139 @@ def __init__(self, api_key, webserver_host="https://citrination.com", suppress_w
...
+            "dataset_ids":
+                dataset_ids
+        }
+
+        failure_message = "Failed to get available columns in dataset(s) {}".format(dataset_ids)
+
+        return self._get_success_json(self._post_json(
+            'datasets/get-available-columns', data, failure_message=failure_message))['data']
+
+    def generate_search_template(self, dataset_ids):
```
sparadiso: This isn't supposed to be user-facing, is it? If not, let's hide it by prepending `__` (aka `__generate_base_search_template`)

citrination_client/views/client.py
```diff
@@ -12,3 +14,139 @@ def __init__(self, api_key, webserver_host="https://citrination.com", suppress_w
...
+            "dataset_ids":
+                dataset_ids
+        }
+
+        failure_message = "Failed to generate a search template from columns in dataset(s) {}".format(dataset_ids)
+
+        return self._get_success_json(self._post_json(
+            'search_templates/from-dataset-ids', data, failure_message=failure_message))['data']
+
+    def prune_search_template(self, extract_as_keys, search_template):
```
sparadiso: Same as above - I think we can hide this from users. `__prune_search_template`?

genfx999: Hide those, and expose a method that invokes those internally?

citrination_client/views/client.py
```diff
@@ -12,3 +14,139 @@ def __init__(self, api_key, webserver_host="https://citrination.com", suppress_w
...
+            "search_template":
+                search_template
+        }
+
+        failure_message = "Failed to prune a search template"
+
+        return self._get_success_json(self._post_json(
+            'search_templates/prune-to-extract-as', data, failure_message=failure_message))['data']
+
+    def create_machine_learning_template(self, search_template, ml_config):
```
sparadiso: Let's not do this here. I'd rather separate concerns - I created `ModelTemplateClient` and `SimplifiedModelTemplateBuilder` stubs for this purpose

citrination_client/views/client.py
```diff
@@ -12,3 +14,139 @@ def __init__(self, api_key, webserver_host="https://citrination.com", suppress_w
...
+            "ml_config":
+                ml_config
+        }
+
+        failure_message = "Failed to create ml template"
+
+        return self._get_success_json(self._post_json(
+            'ml_templates', data, failure_message=failure_message))['data']
+
+    def validate_machine_learning_template(self, ml_template):
```
sparadiso: move to `ModelTemplateClient`

citrination_client/views/client.py
sparadiso: I'd move this into `SimplifiedModelTemplateBuilder`

citrination_client/views/client.py
```diff
@@ -12,3 +14,139 @@ def __init__(self, api_key, webserver_host="https://citrination.com", suppress_w
...
+        failure_message = "ML template validation invoke failed"
+
+        res = self._get_success_json(self._post_json(
+            'ml_templates/validate', data, failure_message=failure_message))['data']
+        if res['valid']:
+            return 'OK'
+        return res['reason']
+
+
+    def create_data_view(self, search_template, ml_template, selected_columns, dataset_ids, name, description):
```
sparadiso: If we have a search template, why do we need dataset_ids and selected columns? Also, total nitpick, I was thinking we could just use `create` for this. That is obviously just UX 🤷‍♂️

citrination_client/views/client.py
```diff
@@ -12,113 +12,7 @@ def __init__(self, api_key, webserver_host="https://citrination.com", suppress_w
...
-
-        res = self._get_success_json(self._post_json(
-            'ml_templates/validate', data, failure_message=failure_message))['data']
-        if res['valid']:
-            return 'OK'
-        return res['reason']
-
-
-    def create_data_view(self, search_template, ml_template, selected_columns, dataset_ids, name, description):
+    def create(self, search_template, ml_template, selected_columns, dataset_ids, name, description):
```
sparadiso: Why isn't this just `create(self, search_template, ml_template, name, description)`?

genfx999: The api for the create requires them

citrination_client/views/search_template/client.py
```diff
@@ -10,5 +10,70 @@ def __init__(self, api_key, webserver_host="https://citrination.com", suppress_w
         members = ["create", "get_available_columns"]
         super(SearchTemplateClient, self).__init__(api_key, webserver_host, members, suppress_warnings, proxies)
 
-    def create(self, ml_template, search_template):
-        pass
+    def create(self, dataset_ids):
```
sparadiso: I think you can just delete this and rename `create_with_extract_as_keys` to `create`

citrination_client/views/client.py
sparadiso: To close the loop on this - that's a bad reason! `rejected` :p

dataset_ids and selected columns is available (in principle) in the search/ml templates.

citrination_client/views/model_template/client.py
```diff
@@ -7,5 +7,47 @@ class ModelTemplateClient(BaseClient):
     """
 
     def __init__(self, api_key, webserver_host="https://citrination.com", suppress_warnings=False, proxies=None):
-        members = ["validate"]
+        members = ["create", "validate"]
         super(ModelTemplateClient, self).__init__(api_key, webserver_host, members, suppress_warnings, proxies)
+
+    def create(self, search_template, ml_config):
```
sparadiso: ModelTemplateClient definitely cannot have a create method that references `ml_config`. ml_config is peculiar to a specific style of creating a template.

citrination_client/views/tests/test_model_template_builder.py
```diff
@@ -79,27 +85,23 @@ def test_workflow():
...
 
         # Validate the template
-        result = client.validate_machine_learning_template(ml_template)
+        result = model_template_client.validate(ml_template)
         assert result == "OK"
 
         # Create the dataview
-        data_view_id = client.create_data_view(search_template, ml_template, available_columns, [1234], 'my view',
-                                               'a test view created by pycc')
+        data_view_id = data_views_client.create(search_template, ml_template, available_columns,
```
sparadiso: Remove `available_columns` and `[1234]` from the test so we can light that up.

citrination_client/views/model_template/client.py
genfx999: That is needed by the endpoint to make the model template, we can internally create the ml config based on something else, but how do we get the descriptor info into the endpoint to make the modeltemplate?

citrination_client/views/client.py
genfx999: Ha!, fair enough.  Per our conversation this morning, removed.

citrination_client/views/model_template/client.py
sparadiso: `ModelTemplateClient` doesn't need to be the only thing that hits an endpoint defined in the `ModelTemplateController`. 

If it makes you feel better, we can refactor the scala code to make the Client<->Controller relationship more clear. I'm focused on getting the API/interface right, then propagating that back to the code for our sanity later.

citrination_client/views/model_template/builders/SimpleModelTemplateBuilder.py
```diff
@@ -0,0 +1,72 @@
...
+class SimpleModelTemplateBuilder(object):
+    """
+    A low dimensional interface for building model templates, based on a basic "ml config" object that specifies
+    descriptor types and basic input/output roles for each column.
+    """
+
+    def __init__(self):
+        self.descriptors = dict()
+
+    def add_real_descriptor(self, column_name, role, upper_bound, lower_bound, units=None):
```
sparadiso: only a true monster would put "upper bound" before "lower bound" in an argument list 👺

citrination_client/views/model_template/builders/SimpleModelTemplateBuilder.py
```diff
@@ -0,0 +1,72 @@
...
+        self.descriptors.update({column_name: descriptor})
+
+    def add_categorical_descriptor(self, column_name, role, descriptorValues, finiteSet):
+        """
+        Add a categorical valued descriptor to the configuration
+
+        :param column_name: Name of the column this descriptor applies to
+        :param role: Input, Output, or LatentVariable
+        :param descriptorValues: An array of categories
+        :param finiteSet: True if the set given is exhaustive, False if not
```
sparadiso: Where does this come from?

citrination_client/views/model_template/client.py
sparadiso: To be clear - my suggestion is that this class _not_ have a create method.

citrination_client/views/client.py
```diff
@@ -0,0 +1,194 @@
+from citrination_client import BaseClient, DataViewStatus, ServiceStatus
+
+
+class DataViewsClient(BaseClient):
+    """
+    Data Views client.
+    """
+
+    def __init__(self, api_key, webserver_host="https://citrination.com", suppress_warnings=False, proxies=None):
+        members = ["create", "update", "delete", "get", "check_predict_status", "predict", "get_predict_result"]
```
sparadiso: Should `predict` be `submit_predict_request`?

citrination_client/views/client.py
```diff
@@ -0,0 +1,194 @@
...
+        result = response["data"]["status"]
+
+        return DataViewStatus(
+            predict=ServiceStatus.from_response_dict(result["predict"]),
+            experimental_design=ServiceStatus.from_response_dict(result["experimental_design"]),
+            data_reports=ServiceStatus.from_response_dict(result["data_reports"]),
+            model_reports=ServiceStatus.from_response_dict(result["model_reports"])
+        )
+
+    def create_ml_configuration(self, search_template, extract_as_keys):
```
sparadiso: I think this should be blocking - poll for results and return the actual config

citrination_client/views/client.py
```diff
@@ -0,0 +1,194 @@
...
+                search_template,
+            "extract-as-keys":
+                extract_as_keys
+        }
+
+        failure_message = "Configuration creation failed"
+        return self._get_success_json(self._post_json(
+            'v1/data_views/builders/simple/default', data, failure_message=failure_message))['id']
+
+    def get_ml_configuration_status(self, job_id):
```
sparadiso: I'd make this private (```__get_ml_...```)

citrination_client/views/client.py
```diff
@@ -0,0 +1,194 @@
...
+        :param dataview_id: The ID of the views
+        :return:
+        """
+        url = 'data_views/{}/retrain'.format(dataview_id)
+        response = self._post_json(url, data={})
+        if response.status_code != 200:
+            raise RuntimeError('Retrain requested ' + str(response.status_code) + ' response: ' + str(response.message))
+        return True
+
+    def predict(self, data_view_id, prediction_source, use_prior, candidates):
```
sparadiso: HMM. Thoughts on polling here and making this call blocking?

citrination_client/views/tests/test_model_template_builder.py
```diff
@@ -0,0 +1,160 @@
...
+        available_columns = search_template_client.get_available_columns([1234])
+        assert len(available_columns) == 524
+
+        # Create a search template from dataset ids
+        search_template = search_template_client.create([1234], available_columns)
+        assert search_template['query'][0]['system'][0]['tags'][0]['category'] == 'General'
+
+        # Create ML configuration
+        dv_builder = DataViewBuilder()
+        dv_builder.add_real_descriptor('Property Melting point', 'Output', 'Infinity', '0')
```
sparadiso: This is actually not crazy. Maybe I take it my prior comment about unifying this back. OR DO BOTH!?

citrination_client/base/response_handling.py
```diff
@@ -29,7 +29,7 @@ def check_for_rate_limiting(response, response_lambda, timeout=1, attempts=0):
 def check_general_success(response, failure_message):
     if response.status_code >= 400:
         raise CitrinationClientError(
-            "{} - Citrination returned {}".format(failure_message, response.status_code)
+            "{} - Citrination returned {}: {}".format(failure_message, response.status_code, response.json()["message"])
```
sparadiso: Although `{"message": ...}` seems standard, it's probably worth catching if it's not available here. I'm pretty sure you can do `response.json().get("message", "Message not available")` or similar.

citrination_client/views/data_view_builder.py
```diff
@@ -0,0 +1,152 @@
...
+        )
+        self.configuration['descriptors'].append(descriptor)
+        if role is not None:
+            self.set_role(descriptor_key, role)
+
+    def add_descriptor(self, descriptor, role=None):
+        """
+        Add a generic descriptor entry.
+
+        :param descriptor: Descriptor dict, see add_*_descriptor methods in this class for examples
```
sparadiso: I feel pretty strongly that we should kill the specific `add_real_descriptor, ...` methods in favor of using citrination_client.models.columns.*. What's the argument against? Also, while we're here, we should add the group_by_key to a growing list and throw an error if you try to add two descriptors with the same name.

genfx999: That is easy enough on the pycc side, FE would need changes to deal with the modified schema going back up/down.  Down in the case of get which returns descriptors, which is what these are.  A hybrid option would be to have builder use them, but format them on request to match current endpoint spec.

PR #236, opened 2019-03-05
[CT-356] Feature/refactor models client

sparadiso:
This PR does 3 things:

1) Removes the old synchronous predict endpoint and replaces it with a polling helper
2) Scrubs the lambda that digs in and pulls top level functionality from subclients to be explicit with those calls (hit a conflict)
3) Consolidates model-related calls to the models client and tucks it into the data views client

### DIFF
--- a/citrination_client/client.py
+++ b/citrination_client/client.py
@@ -10,9 +10,12 @@ Generates a lambda method in a closure such that the
 client and method_name are constant (avoides reassignment
 in the assignment loop in the client class)
 """
+
+
 def _generate_lambda_proxy_method(client, method_name):
     subclient_m = getattr(client, method_name)
     return lambda *args, **kw: subclient_m(*args, **kw)
+
 
 class CitrinationClient(object):
     """
@@ -48,12 +51,12 @@ class CitrinationClient(object):
         self.data = DataClient(api_key, site, suppress_warnings=suppress_warnings, proxies=proxies)
         self.data_views = DataViewsClient(api_key, site, suppress_warnings=suppress_warnings, proxies=proxies)
 
-        clients = [self.models, self.search, self.data, self.data_views]
+        # clients = [self.models, self.search, self.data, self.data_views]
 
-        for client in clients:
-            client_methods = [a for a in dir(client) if not a.startswith('_')]
-            for method in client_methods:
-                setattr(self, method, _generate_lambda_proxy_method(client, method))
+        # for client in clients:
+        #     client_methods = [a for a in dir(client) if not a.startswith('_')]
+        #     for method in client_methods:
+        #         setattr(self, method, _generate_lambda_proxy_method(client, method))
 
     def __repr__(self):
         return "['models', 'search', 'data', 'data_views']"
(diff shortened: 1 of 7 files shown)


citrination_client/models/client.py
```diff
@@ -63,16 +66,21 @@ def predict(self, data_view_id, candidates, method="scalar", use_prior=True):
...
-        response_dict = self._get_success_json(
-            self._post_json(routes.data_view_predict(data_view_id), data=body, failure_message=failure_message))
-        candidate_dicts = response_dict["candidates"]
-        return list(
-            map(
-                lambda c: _get_prediction_result_from_candidate(c), candidate_dicts
+
+        uid = self.submit_predict_request(data_view_id, candidates, method, use_prior)
+
+        while self.check_predict_status(data_view_id, uid)['status'] == "running":
```
nequalszero: Isn't this the endpoint where we settled on [Failed, Killed, Accepted, Finished, Processing] as statuses?

citrination_client/models/client.py
```diff
@@ -129,9 +137,54 @@ def _get_predict_body(self, candidates, method="scalar", use_prior=True):
             }
         }
 
+    def submit_predict_request(self, data_view_id, candidates, prediction_source='scalar', use_prior=True):
+        """
+        Submits an async prediction request.
+
+        :param data_view_id: The id returned from create
+        :param candidates: Array of candidates
+        :param prediction_source: 'scalar' or 'from_distribution'
```
nequalszero: `scalar_from_distribution`?

citrination_client/models/client.py
sparadiso: You're commenting on a past commit (this was fixed in 3a9c464)

citrination_client/models/client.py
sparadiso: Ah, thank you! Good catch. I fixed it in `predict` but not here.

citrination_client/models/client.py
nequalszero: ah ok, I was reviewing commit by commit

nequalszero:
Shouldn't this by `trorvis-ci`? 🤔 
image

citrination_client/client.py
```diff
@@ -48,12 +51,12 @@ def __init__(self, api_key=None, site=None, suppress_warnings=False, proxies=Non
         self.data = DataClient(api_key, site, suppress_warnings=suppress_warnings, proxies=proxies)
         self.data_views = DataViewsClient(api_key, site, suppress_warnings=suppress_warnings, proxies=proxies)
 
-        clients = [self.models, self.search, self.data, self.data_views]
+        # clients = [self.models, self.search, self.data, self.data_views]
```
nequalszero: Why's this commented out?

sparadiso: I ran into a conflict with the little lambda injector below.

citrination_client/client.py
```diff
@@ -10,10 +10,13 @@
 client and method_name are constant (avoides reassignment
 in the assignment loop in the client class)
 """
+
+
 def _generate_lambda_proxy_method(client, method_name):
```
nequalszero: you probably want to remove this method if you're removing the dynamic method definitions

PR #243, opened 2019-04-11
Fix for categorical descriptors

sparadiso:
Categorical descriptors are not consistent with the actual data views API. 

Note: If you prefer to fix this server side so that we don't have to inject logic like this into the client, I'm happy to close this @nequalszero.


nequalszero:
image
The API specifies `descriptor_values` in snake case

sparadiso:
Good call @nequalszero

PR #245, opened 2019-04-17
Move descriptor range to positive number space

genfx999:


sparadiso:
This change has been obviated by moving view API tests to the api tests.

PR #249, opened 2019-05-14
Added int range and value constraints.

latture:
This PR adds range and value constraints for integers. This implementation mirrors that of real range and value constraints.


citrination_client/models/design/constraints/tests/test_constraints.py
```diff
@@ -54,6 +67,47 @@ def test_real_range_constraint_validation():
...
+    try:
+        RealRangeConstraint(name="Ingredient count", minimum=minimum, maximum=maximum)
+        assert False, "IntRangeConstraint should require that minimum be less than maximum"
+    except CitrinationClientError:
+        pass
+
+    # Test values must be castable to float
+    minimum = {}
+    maximum = 2
+    try:
```
sparadiso: FYI py.test has a pattern for asserting exceptions are raised: https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions

PR #252, opened 2019-05-20
Apply proxy to S3 data uploads put

genfx999:


genfx999:
Would this fix the malachite data upload problem?

PR #255, opened 2019-05-24
force environment vars to not override proxies in requests

eddotman:
This forces all `requests` calls to use only the given proxies and not let local environment variables override the given proxies. This should fix some rare proxy-related errors.

### DIFF
--- a/citrination_client/base/base_client.py
+++ b/citrination_client/base/base_client.py
@@ -85,7 +85,10 @@ class BaseClient(object):
         """
         headers = self._get_headers(headers)
         response_lambda = (
-            lambda: requests.get(self._get_qualified_route(route), headers=headers, verify=False, proxies=self.proxies)
+            lambda: requests.get(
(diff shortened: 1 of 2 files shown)


eddotman:
Huh -- looks like checks failed. `It worked on my machine` so I'll look into this. It seems like the new keyword arg I'm passing in isn't being recognized.

sparadiso:
@eddotman feel free to bump the requests version

sparadiso:
https://stackoverflow.com/questions/14351915/where-did-requestss-config-option-go

eddotman:
Ok, after some digging, I found this: https://github.com/kennethreitz/requests/blob/master/requests/sessions.py#L690-L717

In the newer versions of `requests`, environment variables cannot override the explicit proxies given, so I believe that means this 'fix' is unneeded (i.e., there's no need to distrust the environment if the expected use case is to pass explicit proxies).

I think this means we can close this PR and delete the branch. cc @sparadiso @genfx999

PR #257, opened 2019-05-29
 Add version property to BaseClient

genfx999:


citrination_client/base/base_client.py
```diff
@@ -158,4 +161,14 @@ def _delete(self, route, headers=None, failure_message=None):
         return self._handle_response(response, failure_message)
 
     def __repr__(self):
-        return "{}".format(self.api_members)
\ No newline at end of file
+        return "{}".format(self.api_members)
+
+    def __VERSION__(self):
```
maxhutch: I'd shy away from defining new dunder (leading and trailing double underscore) methods.  That's not very pythonic.  Nor is ALL CAPS.

PR #262, opened 2019-06-20
Change relative path for local version lookup, added path to conf.py fix sphinx error 

genfx999:
Note: holding on gh-pages until review

### DIFF
--- a/citrination_client/__init__.py
+++ b/citrination_client/__init__.py
@@ -17,7 +17,8 @@ def __get_version():
     try:
         # Try local first, if missing setup.py, then use pkg info
         here = os.path.abspath(os.path.dirname(__file__))
-        with open(os.path.join(here, "../../setup.py")) as fp:
+        print("here:"+here)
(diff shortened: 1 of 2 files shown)


citrination_client/__init__.py
```diff
@@ -17,7 +17,8 @@ def __get_version():
     try:
         # Try local first, if missing setup.py, then use pkg info
         here = os.path.abspath(os.path.dirname(__file__))
-        with open(os.path.join(here, "../../setup.py")) as fp:
+        print("here:"+here)
```
sparadiso: clean?

docs/source/conf.py
```diff
@@ -15,6 +15,7 @@
 import os
 import sys
 sys.path.insert(0, os.path.abspath('../citrination_client'))
+sys.path.insert(0, os.path.abspath('../..'))
```
sparadiso: this is so smelly 😂

genfx999: lol, that answer came from here: https://stackoverflow.com/a/44980548

PR #263, opened 2019-06-20
Feature/async optional

eddotman:
@Imperssonator and I put together some additions that allow you to make some API calls and turn off async (e.g., you can create a data view and have the function wait to return until the initial models are trained and predict is ready).

tl;dr of changes, major to minor:
- Added a base waiting function
- data / models / views call this wait function in some API calls with an optional `is_async=False` arg
- added tests to ensure these non async routes work (involves some new mock routes)
- also added mock routes / tests for view updating (async and non async)
- minor documentation fixes

### DIFF
--- a/citrination_client/base/base_client.py
+++ b/citrination_client/base/base_client.py
@@ -1,5 +1,6 @@
 import requests
 import json
+import time
 
 from citrination_client.util.quote_finder import quote
 from citrination_client.base.response_handling import raise_on_response, check_general_success, check_for_rate_limiting, get_response_json
@@ -56,6 +57,18 @@ class BaseClient(object):
         raise_on_response(response)
         check_general_success(response, failure_message)
         return response
+
+    def _wait_for(self, activity_description, criterion, timeout, check_delay=5):
+        start = time.time()
+        while time.time() - start < timeout:
+            time.sleep(check_delay)
+            try:
+                if criterion():
+                    return
+            except Exception as e:
+                print("Error checking wait criterion: {}".format(e))
+
+        raise RuntimeError("Failed to `{}` in {} seconds".format(activity_description, timeout))
 
     def _get_success_json(self, response):
         return get_response_json(response)
--- a/citrination_client/base/base_client.py
+++ b/citrination_client/base/base_client.py
@@ -68,7 +68,7 @@ class BaseClient(object):
             except Exception as e:
                 print("Error checking wait criterion: {}".format(e))
 
-        raise RuntimeError("Failed to `{}` in {} seconds".format(activity_description, timeout))
+        raise RuntimeWarning("`{}` not yet complete after {} seconds, please check status manually".format(activity_description, timeout))
 
     def _get_success_json(self, response):
         return get_response_json(response)
--- a/citrination_client/data/client.py
+++ b/citrination_client/data/client.py
@@ -79,7 +79,7 @@ class DataClient(BaseClient):
             if not is_async:
                 self._wait_for(
                     "Dataset ingestion finished",
-                    lambda x: self.get_ingest_status(dataset_id) != "Finished",
+                    lambda x: self.get_ingest_status(dataset_id) == "Finished",
                     timeout)
             return upload_result
         elif os.path.isfile(source_path):
@@ -103,7 +103,7 @@ class DataClient(BaseClient):
                     if not is_async:
                         self._wait_for(
                             "Dataset ingestion finished",
-                            lambda x: self.get_ingest_status(dataset_id) != "Finished",
+                            lambda x: self.get_ingest_status(dataset_id) == "Finished",
                             timeout)
                     return upload_result
                 else:
--- a/citrination_client/data/client.py
+++ b/citrination_client/data/client.py
@@ -81,7 +81,7 @@ class DataClient(BaseClient):
             if not is_async:
                 self._wait_for(
                     "Dataset ingestion finished",
-                    lambda _: self.get_ingest_status(dataset_id) == "Finished",
+                    lambda: self.get_ingest_status(dataset_id) == "Finished",
                     timeout)
             return upload_result
         elif os.path.isfile(source_path):
@@ -105,7 +105,7 @@ class DataClient(BaseClient):
                     if not is_async:
                         self._wait_for(
                             "Dataset ingestion finished",
-                            lambda _: self.get_ingest_status(dataset_id) == "Finished",
+                            lambda: self.get_ingest_status(dataset_id) == "Finished",
                             timeout)
                     return upload_result
                 else:
--- a/citrination_client/data/client.py
+++ b/citrination_client/data/client.py
@@ -54,12 +54,14 @@ class DataClient(BaseClient):
         :return: The result of the upload process
         :rtype: :class:`UploadResult`
         """
+
         upload_result = UploadResult()
         source_path = str(source_path)
         if not dest_path:
             dest_path = source_path
         else:
             dest_path = str(dest_path)
+
         if os.path.isdir(source_path):
             for path, subdirs, files in os.walk(source_path):
                 relative_path = os.path.relpath(path, source_path)
@@ -79,7 +81,7 @@ class DataClient(BaseClient):
             if not is_async:
                 self._wait_for(
                     "Dataset ingestion finished",
-                    lambda x: self.get_ingest_status(dataset_id) == "Finished",
+                    lambda _: self.get_ingest_status(dataset_id) == "Finished",
                     timeout)
             return upload_result
         elif os.path.isfile(source_path):
@@ -103,7 +105,7 @@ class DataClient(BaseClient):
                     if not is_async:
                         self._wait_for(
                             "Dataset ingestion finished",
-                            lambda x: self.get_ingest_status(dataset_id) == "Finished",
+                            lambda _: self.get_ingest_status(dataset_id) == "Finished",
                             timeout)
                     return upload_result
                 else:
--- a/citrination_client/data/client.py
+++ b/citrination_client/data/client.py
@@ -39,7 +39,7 @@ class DataClient(BaseClient):
         ]
         super(DataClient, self).__init__(api_key, host, members, suppress_warnings, proxies)
 
-    def upload(self, dataset_id, source_path, dest_path=None):
+    def upload(self, dataset_id, source_path, dest_path=None, is_async=True, timeout=30):
         """
         Upload a file, specifying source and dest paths a file (acts as the scp command).asdfasdf
 
@@ -47,6 +47,10 @@ class DataClient(BaseClient):
         :type source_path: str
         :param dest_path: The path to the file where the contents of the upload will be written (on the dest host)
         :type dest_path: str
+        :param is_async: Whether or not to make this call asynchronously
+        :type is_async: bool
+        :param timeout: Number of seconds to wait, if not async
+        :type timeout: int
         :return: The result of the upload process
         :rtype: :class:`UploadResult`
         """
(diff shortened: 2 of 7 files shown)


eddotman:
Some things that I might quickly change, pending some feedback:

[ ] Move large mock JSON response to a file?
[ ] Reduce copy/paste code in new tests?

Also, I believe this is non-backwards-breaking, so it's a minor (?) version bump?

citrination_client/base/base_client.py
```diff
@@ -57,6 +58,18 @@ def _handle_response(self, response, failure_message=DEFAULT_FAILURE_MESSAGE):
...
 
+    def _wait_for(self, activity_description, criterion, timeout, check_delay=5):
+        start = time.time()
+        while time.time() - start < timeout:
+            time.sleep(check_delay)
+            try:
+                if criterion():
+                    return
+            except Exception as e:
+                print("Error checking wait criterion: {}".format(e))
```
sparadiso: Can you Warn instead of print to avoid polluting someone's stdout?

citrination_client/base/base_client.py
```diff
@@ -57,6 +58,18 @@ def _handle_response(self, response, failure_message=DEFAULT_FAILURE_MESSAGE):
...
+
+        raise RuntimeWarning("`{}` not yet complete after {} seconds, please check status manually".format(activity_description, timeout))
```
sparadiso: Isn't this an error state (`TimeoutError`)?

citrination_client/data/client.py
```diff
@@ -39,23 +39,29 @@ def __init__(self, api_key, host="https://citrination.com", suppress_warnings=Fa
         ]
         super(DataClient, self).__init__(api_key, host, members, suppress_warnings, proxies)
 
-    def upload(self, dataset_id, source_path, dest_path=None):
+    def upload(self, dataset_id, source_path, dest_path=None, is_async=True, timeout=30):
```
sparadiso: could we call this flag `block_until_complete` or similar? It just isn't intuitive quite intuitive to me. `is_async=True` implies that I'll get back a UUID from this (and changing the return type based on an input parameter value seems very strange).

citrination_client/models/client.py
```diff
@@ -86,16 +87,27 @@ def predict(self, data_view_id, candidates, method="scalar", use_prior=True):
                 "Prediction failed: UID={}, result={}".format(uid, result["status"])
             )
 
-    def retrain(self, dataview_id):
+    def retrain(self, data_view_id, is_async=True, timeout=TRAIN_TIMEOUT):
```
sparadiso: could we call this flag `block_until_complete` or similar?

citrination_client/models/client.py
```diff
@@ -86,16 +87,27 @@ def predict(self, data_view_id, candidates, method="scalar", use_prior=True):
...
-        url = 'data_views/{}/retrain'.format(dataview_id)
+        url = 'data_views/{}/retrain'.format(data_view_id)
         response = self._post_json(url, data={})
         if response.status_code != requests.codes.ok:
             raise RuntimeError('Retrain requested ' + str(response.status_code) + ' response: ' + str(response.message))
+        
+        if not is_async:
+            self._wait_for(
+                "Predict services ready",
+                lambda: self.get_data_view_service_status(data_view_id).predict.ready,
```
sparadiso: This isn't necessarily correct. Predict status will be `ready=True` if you have a view that you trained previously even if you add new data and call `retrain` again. 

I dealt with this in the stress tests by first waiting for `progress<100` _then_ waiting for `progress = 100`.

citrination_client/models/tests/test_models_client.py
```diff
@@ -73,6 +73,16 @@ def test_retrain():
     assert resp == True
 
 
+@pytest.mark.skipif(environ['CITRINATION_SITE'] != "https://citrination.com",
+                    reason="Retrain tests only supported on open citrination")
+def test_retrain_non_async():
+    """
+    Test that we can trigger a retrain
+    """
+    resp = client.retrain("5909", is_async=False)
```
sparadiso: Is this actually testing anything?

citrination_client/views/client.py
```diff
@@ -46,13 +51,19 @@ def create(self, configuration, name, description):
             'v1/data_views', data, failure_message=failure_message))
         data_view_id = result['data']['id']
 
+        if not is_async:
+            self._wait_for(
+                "Predict services ready",
```
sparadiso: It's not clear that me that this is desirable or obvious behavior. It makes me wonder if the view is actually finished while model reports are still being built. What happens if the view can't train any models? Is that a timeout?

citrination_client/views/tests/test_model_template_builder.py
```diff
@@ -84,6 +89,110 @@ def test_workflow():
         data_view_id = data_views_client.create(dv_config, "my view", "my description")
         assert data_view_id == 555
 
+        # Update an ML template
+        data_views_client.update("555", dv_config, "my view", "my description")
+
+
+def test_workflow_non_async():
```
sparadiso: This test seems mostly redundant with the test above and also doesn't appear to meaningfully test the sync v.s. async flag? I might be misreading it, though 😞

citrination_client/data/client.py
eddotman: Great point -- I think a lot of the confusion stemmed from the weird naming I picked. The boolean flags actually work the reverse of how I think you were reading it. Using your name makes it very obvious when the calls are blocking or not.

citrination_client/models/tests/test_models_client.py
eddotman: Yes -- the wait happens when the flag is `False` but this is fixed to work the other way around now.

citrination_client/views/tests/test_model_template_builder.py
eddotman: I think this is the flags being reversed again -- it's changed now to be more clear (i.e., `True` means "wait")

citrination_client/views/client.py
eddotman: Agreed that this is a little prescriptive in kind of an odd way. It does align with lots of workflows we've seen in data science though -- often, you'll derisk a view by building it and watching to see what happens in the UI (or polling status in API). But once that view works / trains, you may want to build ~10 other views that are slightly different and you have high confidence will work (e.g. slightly different ML configs w/ varying inputs/outputs), and typically these subsequent views will work fine -- so this is just a convenience flag so that when the function returns, you know that you can load the view and look at model reports.

eddotman:
Yeah I agree that these flags are a little bit weird from an API design perspective. They are heavily influenced by workflows we've seen internally in data science and also w/ external partners (e.g., customers). IMO, the existing status checking API is fine if you want maximum flexibility, but if you want to quickly run + re-run jupyter notebooks with API calls, and know that everything is "done" when the code finishes (from the perspective of data upload -> ingest -> train -> look at model reports), then these new flags are meant to support that specific workflow. Since the flags are all off by default, the hope is that it won't confuse anyone who isn't specifically looking for them.

eddotman:
Whew ok -- all checks passing now. I had to make some weird mock routes to get some of the progress checks to work, so happy to hear feedback on that. Also happy to hear more high level thoughts on the setup as a whole, given @sparadiso 's thoughts about this functionality being a little strange as a flag in some functions.

citrination_client/views/client.py
sparadiso: Would it significantly change the workflow to give users a utility, so that they can be explicit about what behavior they want? Something like:

```python
models_client.retrain(view_id)
models_client.wait_for_retrain(view_id)
print("retrain finished - updated models are ready")

models_client.wait_for_model_reports(view_id)
print("updated model reports ready")
```

citrination_client/base/base_client.py
```diff
@@ -57,6 +58,18 @@ def _handle_response(self, response, failure_message=DEFAULT_FAILURE_MESSAGE):
...
 
+    def _wait_for(self, activity_description, criterion, timeout, check_delay=5):
+        start = time.time()
+        while time.time() - start < timeout:
+            time.sleep(check_delay)
+            try:
+                if criterion():
+                    return
+            except Exception as e:
+                raise RuntimeWarning("Error checking wait criterion: {}".format(e))
```
sparadiso: Are you sure you meant to raise RuntimeWarning here? I'm fairly sure this will exit on exceptions (note - these exceptions can be blips/timeouts/etc.).

Suggestion: `warnings.warn("Error checking wait criterion: {}".format(e)", RuntimeWarning))`

Imperssonator: This has changed from the last time I saw it...
The basic idea was to copy the `_wait_for()` from citrination-workflow-tests:
```
start = time.time()
        while time.time() - start < timeout:
            time.sleep(check_delay)
            try:
                if criterion():
                    return
            except Exception as e:
                print("Error checking wait criterion: {}".format(e))

        raise RuntimeError("Failed to `{}` in {} seconds".format(activity_description, timeout))
```
But this would throw an Error even if it was just a timeout, and the process could still succeed. Basically I didn't want the functions using `_wait_for()` to throw an error when it's just a timeout. I don't want a timeout to block the `return` of an `upload_result`, for example. So yes, `warnings.warn` would be good. But this is now making me reconsider how these are implemented in general.

eddotman: Yeah I probably should have asked for more clarification before changing this. There was a request to make this throw an error/exception, so I changed it to that, and another (unrelated) request to not send more things to stdout (which the current built-in warning function does). 

Agreed that `warnings.warn` makes sense.

citrination_client/views/client.py
Imperssonator: A lot of my motivation for this comes from writing the SL demo, where `upload`, `create_view`, `retrain` and `design` are happening in a rapid loop, and where function calls need to be synchronous (i.e. next cell or next iteration shouldn't run until previous process is done - very similar to a workflow-test).

Also, I agree with what @eddotman is saying - this would help mitigate the part of DS workflows where we create a view and then stare at the web UI until reports are ready, or upload a dataset then watch ingest progress on the Web UI. But we're smart enough to add our own `waits`, for the most part I think. This is more for non-power-users and community folks who are just getting started.

I think, with respect to both of the above use cases, we're trying to address the following:
Does it help the user to return a successful `upload_result` when the user can't actually do anything with those uploaded files yet? I understand the usefulness internally, but I can't think of a user workflow that doesn't require [this entire series of checks] before doing anything else(https://github.com/CitrineInformatics/citrination-workflow-tests/blob/master/workflow_tests/stress/heavy_correctness_tests/base/data_view_workflow.py#L66-L97).

Similarly, I'm struggling to think of a workflow in which returning a `data_view_id` after a `create()` call is helpful to the user if the services are ultimately going to fail.

Having written that, though, I see the folly of `_waiting_on` predict services to be _ready_, when they might just fail outright.

Imperssonator: there's two ways this could be implemented:
1) *build the `_wait` into the resource instantiation*: don't return anything from the instantiating calls (`upload` or `create`) until the resource is ready to be used
2) *build the `_wait` into the individual resource usage functions*: allow user to call `retrain`, `predict`, `design` on a View, or try to `get_available_columns` from a Dataset, but have those functions `wait` until they can be used successfully. I know that this is already kind of baked into predict requests and design runs.

sparadiso: To be clear, I'm still not blocking this, but for the sake of the dialogue:

I think where I'm coming from is just a general software bias towards simple, atomic functions as the basic units of work. If you want to support an SL workflow that has some assumptions about use case built in, then I'd create a separate package to support that workflow (it could even be a `SequentialLearningClient` that uses the atomic API + sugar). This way, you retain the flexibility and coherence of the base API while also making your users' lives easier for common workflows.

Imperssonator: I appreciate that, and I don't want to bloat PyCC unnecessarily. An `SL_Client` is a great idea, although it would take longer to put together (and I'm about to go on paternity leave). That would also be a good place to integrate some of the other design wrappers we have.

I'm OK closing without merging, but using what we've started here to build an SL Client in the future. What do you think, @eddotman ?

eddotman: Agreed that an SLClient is a great path forward. I do think that if we close this w/o merging, there is a high risk of forgetting to come back to it. 

Can we leave this PR open (probably for a while) and block closing it on the creation of a `SLClient` some time later this summer?

mcgalcode:
I don't have much to add to the existing conversation here, but I also want to acknowledge that I have looked at these changes.

mcgalcode:
@eddotman I'm removing myself as a reviewer from this PR to clean up my PR reminders - my last comment still stands though.

PR #277, opened 2019-09-23
Feature/reformatting and 404 handling

nequalszero:


nequalszero:
reopening from citrine fork

PR #280, opened 2019-10-09
Custom ingester support

nequalszero:
Create IngesterList and Ingester classes

      These IngesterList class will serve to abstract away the list of
      ingester dicts that are available on a Citrination deployment, allowing
      consumers to find the ingester they want to use either via #find_by_id or
      one or more #where calls.

      The Ingester class serves as an abstraction of an ingester dict. It provides
      methods for returning a copy of itself, finding an argument by name, and
      serialization for submission to Citrination.

Create IngestClient

      This client will be leveraged by the DataClient in order to retrieve
      available ingesters, validate ingesters before submission, and to
      handle ingest submissions.

Add custom ingestion endpoints to the DataClient

      Add instantiation of the IngestClient to the DataClient's
      initialization method.

      Add #list_ingesters method, which just invokes IngestClient#list_ingesters

      Add #upload_with_ingester method, which calls IngestClient#validate_ingester
      before making an ordinary file upload request via DataClient#upload, followed
      up by a call to IngestClient#submit to tell Citrination to use a particular
      ingester on the uploaded file.

### DIFF
--- a/citrination_client/data/client.py
+++ b/citrination_client/data/client.py
@@ -81,7 +81,10 @@ class DataClient(BaseClient):
                             dataset_id, current_source_path, current_dest_path
                         )
                         if file_upload_result.successful():
-                            upload_result.add_success(current_source_path)
+                            file_id = file_upload_result.successes[0]['id']
+                            upload_result.add_success(
+                                current_source_path, file_id, current_dest_path
+                            )
                         else:
                             upload_result.add_failure(
                                 current_source_path, "Upload failure"
@@ -126,7 +129,9 @@ class DataClient(BaseClient):
                     self._post_json(
                         routes.update_file(file_data['file_id']), data = data
                     )
-                    upload_result.add_success(source_path)
+                    upload_result.add_success(
+                        source_path, file_data['file_id'], path_data["dest_path"]
+                    )
                     return upload_result
                 else:
                     raise CitrinationClientError(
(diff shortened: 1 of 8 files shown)


citrination_client/data/ingest/ingester.py
```diff
@@ -0,0 +1,104 @@
...
+            "id": self.id,
+            "name": self.name,
+            "namespace": self.namespace,
+            "version": self.version,
+            "display_name": self.display_name,
+            "description": self.description,
+            "arguments": self.arguments
+        }
+
+    # def validate_inputs(self, inputs):
```
mcgalcode: rm ?

citrination_client/data/client.py
```diff
@@ -141,6 +145,45 @@ def upload(self, dataset_id, source_path, dest_path=None):
...
+        self._ingest.validate_ingester(ingester, ingester_arguments)
+        file_upload_result = self.upload(dataset_id, source_path, dest_path)
+
+        if len(file_upload_result.successes) == 1:
+            file_path = file_upload_result.successes[0]["dest_path"]
+            self._ingest.submit(dataset_id, file_path, ingester, ingester_arguments)
+
+        return file_upload_result
+
+    def list_ingesters(self):
```
mcgalcode: Why do you not just access the child client directly?

nequalszero: I was originally going to have direct interactions with the `IngestClient`, but I was unable to encompass all of the upload logic within `IngestClient#submit` due to its need to leverage `DataClient#upload`. I talked with @genfx999 I think, and we concluded that we might as well have ingester related actions go through the `DataClient` rather than have users use the `IngestClient` solely for its `list_ingesters` method.

citrination_client/data/ingest/ingester.py
nequalszero: yes 👍

PR #299, opened 2019-11-11
Draft of add_relations on dataview builder

genfx999:


citrination_client/views/data_view_builder.py
```diff
@@ -58,6 +63,61 @@ def add_raw_descriptor(self, descriptor):
...
+        elif isinstance(inputs, list):
+            if len(inputs) == 0:
+                raise ValueError("Inputs list must not be empty")
+            relation_obj['inputs'] = inputs
+        else:
+            raise ValueError("Unexpected type for inputs, expecting either a string or list of strings")
+
+        if isinstance(outputs, str):
+            relation_obj['outputs'] = [outputs]
+        elif isinstance(outputs, list):
```
EricLundberg: Can a single relation support multiple outputs?

genfx999: Ha! it makes no sense for lolo relations, for material law stuff it does, but that does not apply here.  Ill update to single value.

citrination_client/views/data_view_builder.py
```diff
@@ -63,13 +63,13 @@ def add_raw_descriptor(self, descriptor):
         """
         self.configuration['descriptors'].append(descriptor)
 
-    def add_relation(self, inputs, outputs, relation_type='lolo', options=None):
+    def add_relation(self, inputs, output, relation_type='lolo', options=None):
```
EricLundberg: Do you run any checks to ensure the input(s) and output are in the dataview builder's list of descriptors

genfx999: Im not currently, but that is a good check.  Ill add it.

citrination_client/views/data_view_builder.py
```diff
@@ -58,6 +61,69 @@ def add_raw_descriptor(self, descriptor):
...
+        else:
+            raise ValueError("Unexpected type for inputs, expecting either a string or list of strings")
+
+        if isinstance(output, str):
+            relation_obj['output'] = output
+        else:
+            raise ValueError("Unexpected type for output, expecting a string")
+
+        if relation_type != 'lolo':
+            raise ValueError("Currently, \'lolo\' is the only allowed relation type")
```
nequalszero: minor comment - don't think you need backslashes for the single quotes

citrination_client/views/data_view_builder.py
```diff
@@ -58,6 +61,69 @@ def add_raw_descriptor(self, descriptor):
...
+
+        relation_obj['type'] = relation_type
+
+        # check for duplicate
+        existing_relations = self.configuration['relations']
+        for pos, ele in enumerate(existing_relations):
+            if existing_relations[pos]['inputs'] == relation_obj['inputs'] and \
```
nequalszero: since `inputs` can potentially be an array, you might want to sort the arrays or cast them to a set for comparison

citrination_client/views/data_view_builder.py
```diff
@@ -58,6 +61,69 @@ def add_raw_descriptor(self, descriptor):
...
+        :param relation_type: Kind of relation, currently only 'lolo' is supported
+        """
+        relation_obj = {}
+
+        if isinstance(inputs, str):
+            relation_obj['inputs'] = [inputs]
+        elif isinstance(inputs, list):
+            if len(inputs) == 0:
+                raise ValueError("Inputs list must not be empty")
+            relation_obj['inputs'] = inputs
```
nequalszero: do you do a check anywhere to ensure that there are no duplicates in the array?

citrination_client/views/data_view_builder.py
```diff
@@ -58,6 +61,69 @@ def add_raw_descriptor(self, descriptor):
...
+            raise ValueError("Unexpected type for output, expecting a string")
+
+        if relation_type != 'lolo':
+            raise ValueError("Currently, \'lolo\' is the only allowed relation type")
+
+        relation_obj['type'] = relation_type
+
+        # check for duplicate
+        existing_relations = self.configuration['relations']
+        for pos, ele in enumerate(existing_relations):
```
nequalszero: minor comment - I don't think you to `enumerate` across `existing_relations`, `existing_relations[pos]` is just `ele`, and `pos` doesn't appear to be used for anything else in this loop

citrination_client/views/data_view_builder.py
```diff
@@ -58,6 +61,69 @@ def add_raw_descriptor(self, descriptor):
...
+                    existing_relations[pos]['output'] == relation_obj['output']:
+                raise ValueError("This relation duplicates an existing relation")
+
+        # check limits
+        if len(existing_relations) > MAX_USER_RELATIONS:
+            raise ValueError("Maximum Relations Reached: Citrination only supports " + str(MAX_USER_RELATIONS) +
+                             " user-defined relations. Please review the existing relations")
+
+        # check inputs exist
+        for input_index, input in enumerate(relation_obj['inputs']):
```
nequalszero: Similar enumeration comment here, `input_index` isn't being used.

citrination_client/views/data_view_builder.py
```diff
@@ -58,6 +61,69 @@ def add_raw_descriptor(self, descriptor):
...
+            found = False
+            for desc_index, desc in enumerate(self.configuration['descriptors']):
```
nequalszero: Similar enumeration comment here, `desc_index` isn't being used.

citrination_client/views/data_view_builder.py
```diff
@@ -58,6 +61,69 @@ def add_raw_descriptor(self, descriptor):
...
+                if desc['descriptor_key'] == input:
+                    found = True
+                    break
+            if not found:
+                raise ValueError("Input " + input + " is not defined as a descriptor")
+
+        found = False
+        for desc_index, desc in enumerate(self.configuration['descriptors']):
```
nequalszero: Similar enumeration comment here, `desc_index` isn't being used.

citrination_client/views/data_view_builder.py
nequalszero: Also `input` is a built-in function in python, would recommend renaming to `relation_input` or something

citrination_client/views/data_view_builder.py
```diff
@@ -58,6 +61,69 @@ def add_raw_descriptor(self, descriptor):
...
+            relation_obj['inputs'] = [inputs]
+        elif isinstance(inputs, list):
+            if len(inputs) == 0:
+                raise ValueError("Inputs list must not be empty")
+            relation_obj['inputs'] = inputs
+        else:
+            raise ValueError("Unexpected type for inputs, expecting either a string or list of strings")
+
+        if isinstance(output, str):
+            relation_obj['output'] = output
```
nequalszero: Do we need to check if the output is equal to the input?

citrination_client/views/data_view_builder.py
nequalszero: I think we check the validity of a descriptor key enough times that it's worth creating a `set` of the descriptor keys before this loop begins. Then we can avoid the nested loop and `found` variable, it would just become `if desc in descriptor_keys_set`

citrination_client/views/tests/test_data_view_builder.py
```diff
@@ -20,3 +20,46 @@ def test_add_descriptor():
...
+
+    # limit exceeded
+    limit_hit = False
+    for x in range(0, 50):
+        try:
+            builder.add_descriptor(RealDescriptor("a" + str(x), lower_bound=-1, upper_bound=1), "Input", True)
+            builder.add_descriptor(RealDescriptor("b" + str(x), lower_bound=-1, upper_bound=1), "Output", True)
+            builder.add_relation('a' + str(x), 'b' + str(x))
+        except ValueError:
+            if x > 0:
```
nequalszero: Is this testing the limit of 30 relations?

citrination_client/views/tests/test_data_view_builder.py
```diff
@@ -20,3 +20,46 @@ def test_add_descriptor():
...
+    builder.add_descriptor(RealDescriptor("b", lower_bound=-1, upper_bound=1), "Output", True)
+    builder.add_descriptor(RealDescriptor("c", lower_bound=-1, upper_bound=1), "Output", True)
+
+    assert not check_exception(lambda: builder.add_relation(123, 123))
+    assert check_exception(lambda: builder.add_relation('a', 'b'))
+    assert check_exception(lambda: builder.add_relation('a', 'c'))
+    assert not check_exception(lambda: builder.add_relation('c', ['d']))
+    assert not check_exception(lambda: builder.add_relation([], 'b'))
+    assert not check_exception(lambda: builder.add_relation(['a'], []))
+
```
nequalszero: if making the validation changes I commented about above, should add

`assert not check_exception(lambda: builder.add_relation('a', 'a'))` - output != input
`assert not check_exception(lambda: builder.add_relation(['a', 'a'], 'b'))` - no duplicate inputs

citrination_client/views/tests/test_data_view_builder.py
genfx999: yessir

citrination_client/views/data_view_builder.py
genfx999: Added this check also.

citrination_client/views/data_view_builder.py
genfx999: Set works great for this.  Updated on latest commit.

Issue #235, opened 2019-03-01
target.py docstring does not reflect all possible values for objective param


matSciMalcolm:
The Target object may now accept a specific value in addition to "Min" or "Max", but this is not expressed within the existing docstring.

If the following could be added to the `target.py` docstring `(either "Max" or "Min")` ->`(either "Max", "Min", or a specific value for the property)` I think it would help users better debug/build tools.
```
 """
    The optimization target for a design run. Consists of
    the name of the output column to optimize and the objective
    (either "Max", "Min", or a specific value for the property)
    """

    def __init__(self, name, objective):
        """
        Constructor.

        :param name: The name of the target output column
        :type name: str
        :param objective: The optimization objective; either "Min",
                "Max", or a specific value for the property
        :type objective: str
        """
```

PR #301, opened 2020-02-04
Create AdvancedDataViewBuilder

nequalszero:
Create BaseDataViewBuilder and AdvancedDataViewBuilder

      Refactor the re-usable components of DataViewBuilder into the
      BaseDataViewBuilder.
      Create the AdvancedDataViewBuilder for configuring views with relations.

Move relation tests into advanced data view builder test file


citrination_client/models/design/target.py
```diff
@@ -5,7 +5,7 @@ class Target(object):
     """
     The optimization target for a design run. Consists of
     the name of the output column to optimize and the objective
-    (either "Max" or "Min")
+    (either "Max" or "Min", or a scalar value (such as "5.0"))
```
nequalszero: found an open issue that referenced this docstring inaccuracy https://github.com/CitrineInformatics/python-citrination-client/issues/235

PR #308, opened 2020-04-07
Feature/ct 923 builder formulation support

genfx999:


citrination_client/views/base_data_view_builder.py
```diff
@@ -64,6 +69,27 @@ def get_dagre_graph(self):
             "edges": edges
         }
 
+    def add_formulation_descriptor(self, descriptor, dataview_client):
```
nequalszero, line 72: Given this method is located on the `BaseDataViewBuilder` class, we'll probably want to have it handle adding `roles` in the event that the child class is the `DataViewBuilder` and not the `AdvancedDataViewBuilder`.

genfx999: roger that, its in the queue

PR #309, opened 2020-04-09
Client side changes to support class probabilities

genfx999:


genfx999:
FYI - Will hold since public has new thorin (needed for this change and associated integration test)

citrination_client/models/client.py
```diff
@@ -87,8 +87,8 @@ def predict(self, data_view_id, candidates, method="scalar", use_prior=True):
         result = self.check_predict_status(data_view_id, uid)
         if result["status"] == "Finished":
 
-            paired = zip(result["results"]["candidates"], result["results"]["loss"])
-            prediction_result_format = [{k: (p[0][k], p[1][k]) for k in p[0].keys()} for p in paired]
+            paired = zip(result["results"]["candidates"], result["results"]["loss"], result["results"]["classProbabilities"])
+            prediction_result_format = [{k: (p[0][k], p[1][k], p[2].get(k)) for k in p[0].keys()} for p in paired]
```
nequalszero, line 91: Is `result["results"]["classProbabilities"]` a dictionary? Does it need to be `p[2].get(k)` or can it just be `p[2][k]` like the candidate and loss are referenced?

genfx999: I think it could be p[2][k] vs p[2].get(k).  It is just a dictionary.

PR #311, opened 2020-04-22
Added docs for formulations setup

genfx999:


docs/source/tutorial/view_examples.rst
```diff
@@ -152,3 +152,20 @@ intended to be passed to the ``plot`` function from the ``dagre_py.core`` module
...
+
+``add_formulation_descriptor`` will use the passed in data views client to automatically add the shares, component type
+and the name descriptors.  This requires that the dataset_ids have already been set.
+
+.. literalinclude:: /code_samples/views/data_view_builder_formulation_desc.py
+
+Ingredient Properties in AdvancedDataViewBuilder
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+When defining ingredient level properties using the ``AdvancedDataViewBuilder``, you will need to define relationships
+with the ingredient property as inputs to the formulation descriptor itself.
```
nequalszero, line 169: Is there only ever one relation that needs to be defined involving ingredient properties? If so, would `you will need to define relationships with the ingredient property as inputs to the formulation descriptor itself` be more accurate as `you will need to define a relation with all ingredient properties as inputs to the formulation descriptor itself`?  The motivation there being to clarify that it's a single relation to be added, not multiple relations, and that there can be more than one ingredient property.

PR #314, opened 2020-04-30
Add 'ingest log' retrieval route to PyCC

dannycodes:
https://citrine.atlassian.net/browse/CT-1191


dannycodes:
@genfx999 AFAIK there are no tests for PyCC?

genfx999:
There are some tests, for ingestion stuff look in citrination_client/data/tests folder

dannycodes:
@genfx999 Added a (rather trivial) test here.

Wasn't sure about e2e here. I think to do that I'd have to be able to idempotently create a dataset and then do the same for a file before reading the logs.. seems like overkill all things considered.

